PR #1, opened 2020-11-10
Initial merge from original drewandersonnz repo

drewandersonnz:
Multiple people have committed to this repo, this PR is to bring the commit history of multiple contributors forward.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,44 @@
+#FROM ocm-container
+FROM fedora:latest
+
+ENV I_AM_IN_CONTAINER="I-am-in-container"
+
+RUN yum -y install \
+    bash-completion \
+    findutils \
+    fzf \
+    git \
+    golang \
+    jq \
+    make \
+    procps-ng \
+    rsync \
+    sshuttle \
+    vim-enhanced \
+    wget \
+    && yum clean all;
+
+ADD ./container-setup /container-setup
+
+WORKDIR /container-setup
+
+ARG osv4client=openshift-client-linux-4.3.12.tar.gz
+ARG rosaversion=v0.0.16
+ARG awsclient=awscli-exe-linux-x86_64.zip
+ARG osdctlversion=v0.2.0
+ARG veleroversion=v1.5.1
+
+RUN ./install/install-rosa.sh
+RUN ./install/install-ocm.sh
+RUN ./install/install-oc.sh
+RUN ./install/install-aws.sh
+RUN ./install/install-kube_ps1.sh
+RUN ./install/install-osdctl.sh
+RUN ./install/install-velero.sh
+RUN ./install/install-utils.sh
+
+RUN cat /container-setup/install/bashrc_supplement.sh >> ~/.bashrc
+
+RUN rm -rf /container-setup
+
+WORKDIR /root
--- a/container-setup/install/install-aws.sh
+++ b/container-setup/install/install-aws.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+#export awsclient=awscli-exe-linux-x86_64.zip
+
+mkdir /usr/local/aws;
+pushd /usr/local/aws;
+wget -q https://awscli.amazonaws.com/${awsclient}
+unzip ${awsclient}
+rm ${awsclient}
+./aws/install;
+popd;
--- a/container-setup/install/install-kube_ps1.sh
+++ b/container-setup/install/install-kube_ps1.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+mkdir /usr/local/kube_ps1;
+pushd /usr/local/kube_ps1;
+wget -q https://raw.githubusercontent.com/drewandersonnz/kube-ps1/master/kube-ps1.sh;
+popd;
--- a/container-setup/install/install-oc.sh
+++ b/container-setup/install/install-oc.sh
@@ -1,0 +1,20 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+#export osv4client=openshift-client-linux-4.3.5.tar.gz
+
+# https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/
+mkdir /usr/local/oc;
+pushd /usr/local/oc;
+wget -q https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/${osv4client};
+tar xzvf ${osv4client};
+rm ${osv4client};
+ln -s /usr/local/oc/oc /usr/local/bin/oc;
+oc completion bash >  /etc/bash_completion.d/oc
+popd;
--- a/container-setup/install/install-osdctl.sh
+++ b/container-setup/install/install-osdctl.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+#export osdctlversion=v0.1.0
+
+mkdir /usr/local/osdctl;
+pushd /usr/local/osdctl;
+wget -q https://github.com/openshift/osd-utils-cli/releases/download/${osdctlversion}/osdctl-linux-${osdctlversion};
+mv osdctl{-linux-${osdctlversion},}
+chmod +x osdctl
+ln -s /usr/local/osdctl/osdctl /usr/local/bin/osdctl;
+osdctl completion bash > /etc/bash_completion.d/osdctl;
+popd;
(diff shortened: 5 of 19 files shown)


drewandersonnz:
Should this be one squashed commit? Or maintain the commit history?

@georgettica suggested this: https://stackoverflow.com/questions/7442112/how-to-attribute-a-single-commit-to-multiple-developers

2uasimojo:
Personally I prefer leaving the individual commits.

karthikperu7:
I would vote for keeping the individual commits too.

container-setup/install/install-kube_ps1.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+mkdir /usr/local/kube_ps1;
+pushd /usr/local/kube_ps1;
+wget -q https://raw.githubusercontent.com/drewandersonnz/kube-ps1/master/kube-ps1.sh;
```
clcollins, line 12: This should pull either directly from the upstream kube-ps1, or use a copy forked in this repo, just in case (heaven forbid) Drew decides to leave us someday.

Dockerfile
```diff
@@ -0,0 +1,44 @@
+#FROM ocm-container
+FROM fedora:latest
```
clcollins, line 2: This should probably be a tagged image version that we consciously update when a new version comes out.  I've previously had bad experiences when "latest" changed to point to a newer version and some things broke.  We should ensure that the Dockerfile is in a working tested state, and we can't do that when "latest" can arbitrarily point to a new version.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+ENV I_AM_IN_CONTAINER="I-am-in-container"
+
+RUN yum -y install \
```
clcollins, line 6: `yum` is deprecated/deprecating in current versions of Fedora .  Perhaps change to `dnf`?

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    git \
+    golang \
+    jq \
+    make \
+    procps-ng \
+    rsync \
+    sshuttle \
+    vim-enhanced \
+    wget \
+    && yum clean all;
```
clcollins, line 19: You can save more disk space by removing the yum cache as well:

`&& yum clean all && rm -r /var/cache/yum/`

container-setup/install/install-aws.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: What is the benefit of this? To prevent accidental install?  Could this just be a non-executable file that is set to executable when copied into the container instead?  This just adds extra complexity to all the files and the Dockerfile/is not DRY.

container-setup/install/install-oc.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: Ditto above.

container-setup/install/install-osdctl.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: Ditto above.

container-setup/install/install-rosa.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: Ditto

container-setup/install/install-rosa.sh
```diff
@@ -0,0 +1,28 @@
...
+  echo "must be run in container";
+  exit 1;
+fi
+
+source /container-setup/install/helpers.sh
```
clcollins, line 8: I might be missing something, but is this doing anything?  The only function in there is `remove_coloring()` and I don't see it called/used in here.

container-setup/install/install-utils.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: Ditto

container-setup/install/install-utils.sh
```diff
@@ -0,0 +1,19 @@
...
+  echo "must be run in container";
+  exit 1;
+fi
+
+echo "in container";
+
+mkdir -p ${HOME}/utils
+mkdir -p /etc/profile.d
+echo "export PATH=${PATH}:${HOME}/utils" > /etc/profile.d/localbin.sh
```
clcollins, line 12: If this is running inside a container, designed to be run as root anyway, why not just put the utilities in one of the normal $PATH bin directories?  Or use ~/.local/bin rather than a custom dir?  This file could be replaced with one line in the Dockerfile, maybe.

container-setup/install/install-velero.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+
+if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
```
clcollins, line 3: Ditto

clcollins (approved):
This looks fine to me for an initial commit into the repo. There is nothing internally sensitive included and it's working great.

I added several comments/review notes, but those were mostly as a reminder for things I want to address myself later if someone doesn't have a quick fix.

/ltgm
/approve

rogbas:
Overall it LGTM, we could address the comments from this review in a separate PR

clcollins:
> Overall it LGTM, we could address the comments from this review in a separate PR

@rogbas That was my intention.  Just wanted to get them noted down.

Dockerfile
iamkirkbater: I have mixed feelings about it.  On one hand, we have issues if they update the upstream.  On the other hand, we have to have someone watch the upstream and update our build when necessary.

For a project like this, I think it's safe to stay on latest but be cognizant of that, especially once we get to the point where you won't be building this locally unless you're heavily customizing it or if you're building the release, in which case we'd probably want the latest updated at that point anyway.

With the end goal of providing a quay.io image that the average user of ocm-container can just download and run, as long as we tag _those_ appropriately I'm in the camp of leaving this at :latest, but if I'm the minority I'll concede the argument.

Dockerfile
iamkirkbater: I believe we are going to leave `fedora` as a base image anyway and either move to coreos container linux or centos as our base, but I don't think we've finalized on which we are moving towards.  Good to keep in mind though if something breaks before we make the migration.

Dockerfile
iamkirkbater: Nice, we should add that in!

```suggestion
    && yum clean all \
    && rm -r /var/cache/yum;
```

container-setup/install/install-aws.sh
iamkirkbater: Ooohhh, I really like that idea.  I'll card that up.

container-setup/install/install-rosa.sh
iamkirkbater: You're probably right.  ROSA used to be installed via `go get` I believe and that used to dirty up the terminal (because for whatever reason the red color would persist through the rest of the build).  It should be safe to remove this when we move to including only the binaries instead of the entire source repos.

container-setup/install/install-utils.sh
iamkirkbater: Funny you mention this, I actually had to make some edits to this to get `docker exec` to work and auto-login as well as to be able to run commands from outside the container.

I ended up changing the path in the Dockerfile and removing this in another PR that will come up here after this one merges.

Dockerfile
clcollins: I would argue you have to watch something anyway.  You already have versions of the binaries hard-coded, so at the point they need to be upgraded, a human has to increment a number, just like this.

If you were to keep `latest` and switch to using tests to verify, then you need a human watching the build tests, unless PRs fail to build.  I would say as long as each PR is built as part of it's testing, then that is probably the best solution:

* leave as latest
* move other binaries to latest
* make each PR test require an image build

Dockerfile
clcollins: :+1:

container-setup/install/install-kube_ps1.sh
drewandersonnz: I agree with having it not tied to me, it is a fork in my repo so we can check any changes ... I don't want to reply on someone external to our team for a security review of inbound code.

Last thing we want is an attacker gaining access to and submitting dangerous code to a smaller repo that we depend on.

I'm less concerned about other tools in the `github.com/openshift` organisation, as we are in control of that.

Dockerfile
drewandersonnz: There is a consideration moving to automation for image builds and tests. This should alert us to problems as they happen.

Dockerfile
drewandersonnz: I think it might be good to deploy multiple base image versions... `ocm-container:fedora`, `om-container:ubuntu`, `ocm-container:alpine` for example. 

As we use it as an interactive environment, people shouldn't be made to learn tools they're not familiar with.

container-setup/install/install-aws.sh
drewandersonnz: Yes, intention is to block execution on your local machine, in case of accidental execution outside of container. +1, I like the "not executable until required".

container-setup/install/install-kube_ps1.sh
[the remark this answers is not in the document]
drewandersonnz: As above, this fork is a safety-net, we are not using upstream directly to be able to review any code additions. This could be vendored, or cloned to openshift org, or some other thing. But some link through to finally upstream would be good for recognition of that projects contributions, and for being able to stay in sync.

PR #2, opened 2020-11-20
Adds browserless login

iamkirkbater:
Adds browserless login and the ability to automatically log into a cluster on container launch.


georgettica:
Got some one thing that may help us until srelogin is in a final form:
Can we pull it from the source? On my local setup I put it as a volume and kept pulling it, though for now taking the lastest should work aswell.

georgettica:
I tried to "merge upstream" the srelogin, and found all we need to so is set one of 4 env vars to make it work without modifications

see closed PR https://github.com/mbarnes/work-environment/pull/3#issue-525330571

iamkirkbater:
Nice, thanks for the tip @georgettica.  I've updated the cluster-login script to what's current as of 10 minutes ago and updated ocm-container.sh to reflect the USER env var (I still want to give flexibility for those who's OCM username might be different from their kerberos username.)

georgettica:
which is mine aswell, you can always do
```
$ USER=aaa python3 -c 'import getpass; print(getpass.getuser())'
aaa
```

and call the srelogin with that argument

iamkirkbater:
Right.  If it changed a lot, or on a per-call basis you could do that.  But setting it as a config var inside your env.source once is kind of a set-it-and-forget-it-until-you-have-to-set-this-up-again-and-then-you're-confused-because-nothing's-working-the-way-it-used-to approach.

georgettica:
hahha, oki I am fine with that

georgettica:
@drewandersonnz should we do something else before merging?

maybe a knowledge sharing on the internal setup of the configs?

ocm-container.sh
```diff
@@ -70,14 +71,39 @@ then
...
+fi
+
 ### start container
 ${CONTAINER_SUBSYS} run -it --rm --privileged \
--e "OCM_URL=${OCM_URL}" \
+-e "OCM_URL" \
+-e "USER" \
 -e "SSH_AUTH_SOCK=/tmp/ssh.sock" \
+-e "KRB5CCNAME=/tmp/krb5cc" \
+-e "OFFLINE_ACCESS_TOKEN" \
```
iamkirkbater, line 100: These need to be set here so that this can become scriptable, as ~.bashrc is not run when the container inits without a TTY.

georgettica (approved):

PR #4, opened 2020-11-21
allow sudo prompt init

clcollins:
This adds a prompt to input the sudo password when symlinking the ocm-container script into /usr/local/bin.  This allows the rest of the init script itself to be run without sudo privileges, and is more clear about what is and isn't being done as root.
    
### DIFF
--- a/init.sh
+++ b/init.sh
@@ -44,8 +44,8 @@ else
   init_new_config
 fi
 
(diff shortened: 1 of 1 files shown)


georgettica:
/lgtm
Didn't remember I needed root to install even, thanks!

PR #5, opened 2020-11-22
feat(startup): remove unused variable

georgettica:
as the macos 'ssh-agent' works without issues,
there is no need for runnig ssh-agent before the command


drewandersonnz:
On linux environments this variable was empty, therefore had no effect. Meaning, +1 to merge based on a Fedora linux execution environment.

@iamkirkbater can you confirm it still works as planned on macos? Can you still use your ssh-agent/keyring/whatever to connect outbound with this change?

PR #6, opened 2020-11-22
docs(readme): shorten environment names

georgettica:
as a feat in ocm-cli was introduced to work with these, I would suggest to adopt them


drewandersonnz:
+1 from me, maybe could also add a docs link to the code location where these are defined in `ocm`?

iamkirkbater:
This works for me.

PR #11, opened 2020-12-23
Fix PS1 so that bash new lines work consistently

karthikperu7:
This has been bugging me ever since I started using ocm-container.

When you are in a container (bash) session, if your commands are really long, to the point that it extends more than the width of the current line, then bash starts overwriting your current line instead of starting in a new line.

Similar to this issue in [stack overflow](https://unix.stackexchange.com/questions/49891/bash-overwrites-the-first-line-ps1-bash-prompt) .

The fix is to have the colour code to start and end with `\[` and `\]` so that bash does not count that towards the number of characters before it starts a new line.

I tried to fix it with keeping the bash function but in the end this is what I came up with, which works.

Hope this makes sense!

### DIFF
--- a/container-setup/utils/bashrc_supplement.sh
+++ b/container-setup/utils/bashrc_supplement.sh
@@ -4,7 +4,8 @@ source /usr/local/kube_ps1/kube-ps1.sh
 
 ## Set Defaults
 export EDITOR=vim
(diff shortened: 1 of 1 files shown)


georgettica:
tested and it seems to work. 
/lgtm
/approve

PR #14, opened 2021-01-20
If OCM_CONTAINER_KERBEROS_USER is already set, prefer it

lisa:
Prefer `OCM_CONTAINER_KERBEROS_USER` if it is already set in the environment, otherwise fall back to `$OCM_USER` (if set), and then finally fall back to `$(whoami)`

This helps in cases where the Kerberos user is different from the OCM user, for whatever reasons.

### DIFF
--- a/env.source.sample
+++ b/env.source.sample
@@ -23,7 +23,7 @@ export OCM_URL=${OCM_URL:-}
 export OCM_USER=${OCM_USER:-your_user}
 
 ### Your kerberos username (without domain) if different than your OCM_USER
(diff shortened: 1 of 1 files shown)


drewandersonnz:
/lgtm

@iamkirkbater can you confirm? I don't have kerb on my local... 

I want to merge because it looks so easy, but @iamkirkbater is one of the experts on this :-P

iamkirkbater:
/lgtm  

This looks fine to me.  I don't think it should cause any issues.

PR #16, opened 2021-01-21
Use the refresh token directly for ocm login

feichashao:
We can use the refresh token to login ocm directly, so that the token won't expire (too soon) for ocm commands.

.ocm.json will look like:
~~~
cat .ocm.json 
{
  "access_token": "<masked>",
  "client_id": "cloud-services",
  "refresh_token": "<masked>",
  "scopes": [
    "openid"
  ],
  "token_url": "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
  "url": "https://api.stage.openshift.com"
}
~~~

And every time we run ocm command, it will refresh the ocm access token automatically if expired.

Or, is there any reason we don't do it that way? Security?

### DIFF
--- a/container-setup/utils/bin/ocm-login
+++ b/container-setup/utils/bin/ocm-login
@@ -20,15 +20,7 @@ then
   OCM_URL="https://api.openshift.com"
 fi
 
-TOKEN=$(curl \
---silent \
---data-urlencode "grant_type=refresh_token" \
---data-urlencode "client_id=cloud-services" \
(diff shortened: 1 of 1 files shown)


georgettica:
I'll leave @drewandersonnz to comment on that one

georgettica (commented):
Not much I can say about it :/

drewandersonnz:
Tested, works for me.

/lgtm
/approve

PR #17, opened 2021-01-22
Replaces ocm install via go get with release binary

clcollins:
As of the creation of this PR, building the container image fails attempting to `go get` ocm-cli. This PR switches the ocm install to use the latest release binary, rather than the current main branch code, akin to using a container image tag rather than `:latest`.

### DIFF
--- a/container-setup/install/install-ocm.sh
+++ b/container-setup/install/install-ocm.sh
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+set -o errexit
+
 if [ "$I_AM_IN_CONTAINER" != "I-am-in-container" ]; then
   echo "must be run in container";
   exit 1;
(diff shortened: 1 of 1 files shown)


container-setup/install/install-ocm.sh
```diff
@@ -8,6 +10,19 @@ fi
 echo "in container";
 source /container-setup/install/helpers.sh
 
-remove_coloring go get -v -u github.com/openshift-online/ocm-cli/cmd/ocm;
-ln -s /root/go/bin/ocm /usr/local/bin/ocm;
+LATEST_RELEASE="https://github.com/openshift-online/ocm-cli/releases/latest/download/ocm-linux-amd64"
+
+pushd /usr/local/bin/
```
georgettica, line 15: Not opposed to your idea, but usually on the repo we put each tool.in its own folder

container-setup/install/install-ocm.sh
```diff
@@ -8,6 +10,19 @@ fi
...
+echo "Downloading the latest release"
+curl -sSL ${LATEST_RELEASE} -o ocm-linux-amd64
+echo "Validating binary against SHA256 sum"
+curl -sSL ${LATEST_RELEASE}.sha256 | sha256sum --check --status
```
georgettica, line 19: That is something good we might wanna use in other places

georgettica (commented):
Didn't test but it looks good from my side.
If someone tests before me feel free to go ahead.

If you could review my comments, they are not show-stoppers, but might be worth looking at

container-setup/install/install-ocm.sh
clcollins: Amended to install to /usr/local/ocm, symlink to /usr/local/bin/ocm

PR #21, opened 2021-01-27
Allow OCM_URL to default to production

iamkirkbater:
This allows OCM_URL to default to production, but allow the prompt to be changed by exporting a new var within the cluster.  Still allows the env var to be overridden in the container as well.


georgettica:
Lgtm, merging

PR #22, opened 2021-01-28
Set default namespace for velero in env file

karthikperu7:


env.source.sample
```diff
@@ -25,6 +25,10 @@ export OCM_USER=${OCM_USER:-your_user}
 ### Your kerberos username (without domain) if different than your OCM_USER
 export OCM_CONTAINER_KERBEROS_USER=${OCM_CONTAINER_KERBEROS_USER:-${OCM_USER:-$(whoami)}}
 
+### Set the default namespace for velero to avoid using 
+### --namespace=openshift-velero each time
+velero client config set namespace=openshift-velero
```
iamkirkbater, line 30: I like the change, but can we put this in the supplemental bashrc in `container-setup/utils/bashrc_supplement.sh`?  

I worry that adding it here in the .env.source _might_ run into issues when attempting to build or run the container if you don't have velero installed on your host.

iamkirkbater (changes requested):

env.source.sample
karthikperu7: Thanks, fixed now :)

karthikperu7:
Made the default velero ns configurable using env variable in env.sample.source and bashrc_supplement.sh using that env var instead

iamkirkbater (approved):
/lgtm
/approve

Nice :)

drewandersonnz:
@karthikperu7 should this be set in `bashrc`/`env.source` ?? Or should it be in the `install-velero.sh` script?

As this currently stands, the bashrc is broken for everyone who doesn't realise this change needs to be made to their local `env.source`. At the very least you would need to default the variable in `bashrc` or use an `if var is set; velero set config`. 

I'm not sure I'm happy with the location of this code where it is at the moment, can you explain why you chose this location?

iamkirkbater:
Oh, crap, @drewandersonnz is right.  I missed that when I reviewed last night, no more late night code reviews for me.

I wanted to leave this as extensible as possible so that everyone's config can be whatever they want so I asked @karthikperu7 to make the change.  While I think it's a good idea to have a default, I missed that the default should have been in the script itself, and not in the env.source.

I'll make another PR to fix that. Good catch, Drew.

iamkirkbater:
Hmm, although when I stop to think about it, I wonder if it would make sense like you mention, Drew, to put it in install-velero.sh and then have an optional override in the bashrc supplement.  So it will already be set in the container (instead of calling it during container init, it's already present) and if you have another need for another default it could be set via env var and set to that upon container init.

PR #23, opened 2021-01-28
Remove optional FILE prefix from OCM_CONTAINER_KRB5CC_FILE

mrbarge:
This PR removes the `FILE:` prefix from `OCM_CONTAINER_KRB5CC_FILE` if it occurs, leaving just the path (and making the subsequent `-f` test succeed). This change should not impact any users for whom this is not an issue.

**Background**
Some Kerberos setups (including mine) will set a default `KRB5CCNAME` ccache convention of `FILE:/path/to/kerberos/ccache`. As a result, the current incarnation of `ocm-container.sh` does not successfully mount the ccache file into the container.


iamkirkbater:
/lgtm

Tested and looks good to me.

PR #24, opened 2021-01-29
try to one script

reidliu41:
Try to combine the install* scripts into one script.


drewandersonnz:
Thank you for your contribution.

While the code looks well written, this was an intentional decision to have separate install* scripts for maintainability. 

These scripts install projects that are changing fast, and being able to keep changes to separate files helps us to update and test quickly as required.

Thanks again, but I'm going to close this PR.

PR #25, opened 2021-01-29
Updates velero to be overrideable but set during install

iamkirkbater:
From @karthikperu7's PR yesterday I built off of that and the feedback that @drewandersonnz had to initialize velero with a sane default, but check on container initialization to see if the env is overridden and then reconfigure velero from there.


drewandersonnz:
Previous PR: https://github.com/openshift/ocm-container/pull/22

PR #27, opened 2021-02-24
Jaybeeunix better cluster login

jaybeeunix:
(I know cluster-login is going away with backplane, but I wanted to preserve a couple of enhancements.)

This change adds @mbarnes 's useragent fix and my change to allow selecting clusters by any of the OCM-indexed ID fields (namely, display_name, name, id, or  external_id).


georgettica:
I can +1 that as it looks good enough and I still prefer it to backplane
/lgtm
/approve
(if anyone else wants to merge I would be glad)

georgettica:
@clcollins as you are a reviewer can you do the honors?

PR #28, opened 2021-03-09
Adjust `sre-login` to allow arguments  [WAS: Fixes sre-login to use OCM_USER var]

clcollins:
Currently, the `sre-login` script ignores if the `$OCM_USER` variable is set.  This change adjusts the sre-login script to accept multiple arguments:

    usage: sre-login [-h] [-d] -c CLUSTER [-i OCM_IDENTITY]

    required arguments:
      -c      the name or ID of the cluster to log into

    optional arguments:
      -h      show this help message and exit
      -d      debug (bash xtrace)
      -i      OCM Identity to use for cluster login (usually your Kerberos ID)

The added help message and debug will assist new users or folks having issues tracking down where their login problems are occuring, and the OCM identiy argument will allow setting of any ID needed to login to the cluster.

`OCM_IDENTITY` can also be set via ENV VAR, so it doesn't need to be specified manually.

This also sets the stage to add other arguments if necessary, both for use in this script or as options passed to `cluster-login`.

This comes at the expense of needed to pass `-c cluster` rather than just the bare cluster name, but I think this is more clear usage, and less prone to issues.

I'm happy to adjust to accept the old usage as well, or to add a deprecation message, if that's desired.

~~Currently, the `sre-login` script ignores if the `$OCM_USER` variable is set.  This change modifies the script to pass the `-e ${OCM_USER}@redhat.com` flag when calling `cluster-login` if the `$OCM_USER` variable is set.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Welcome! See 'oc help' to get started.
 ```
 
 ### Automatic Login
-We've built in functionality to simplify the cluster login steps.  Now within the contianer you can run `sre-login cluster-id` and it will refresh your ocm login, create a tunnel within the container if necessary, and then log-in to the cluster.
+We've built in functionality to simplify the cluster login steps.  Now within the contianer you can run `sre-login -c cluster-id` and it will refresh your ocm login, create a tunnel within the container if necessary, and then log-in to the cluster.
 
 `sre-login` accepts both a cluster-name or a cluster-id.  If the cluster-name is not unique, it will not ask which one, but display the clusters and exit.
 
--- a/container-setup/utils/bashrc_supplement.sh
+++ b/container-setup/utils/bashrc_supplement.sh
@@ -26,7 +26,7 @@ fi
 
 if [ -n "$INITIAL_CLUSTER_LOGIN" ]
 then
-  sre-login $INITIAL_CLUSTER_LOGIN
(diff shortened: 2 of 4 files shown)


clcollins:
@georgettica I took this a bit further and added real argument handling so any ID could be passed as an argument, and added a help message/usage message and debug.

clcollins:
Confirmed working with automatic/`INITIAL_CLUSTER_LOGIN` as well, using ENV VARS.

georgettica:
❤️ I will test this today to verify

georgettica:
for some reason it doesn't want to work for me now (with your change or without)

if you can pull someone else to verify it works for them (macos user preferably) that would be super! as I know it's an issue you have and can be solved easily with this simple change

iamkirkbater (changes requested):
I'm a bit confused what the addition of `OCM_IDENTITY` adds?  Also, I'm just a smidge worried about backwards compatibility with removing the positional argument and forcing that to change to a flag, is there a reason why we're moving in that direction?

georgettica:
merged a ton of PR's and this had merge conflicts

waiting for someone to review this, so I can merge it

georgettica:
@clcollins is this still required?

clcollins:
With the backplane stuff, this is no longer necessary/useful.

PR #35, opened 2021-04-14
Update README.md

hlipsig:
Build.sh and init.sh should be swapped as init is required to make the dependencies for build.


mrbarge:
Hit this same thing when walking someone else through installing `ocm-container` this week - agree on switching the order.
/lgtm

georgettica:
If you LGTM then it's good enough for me to merge

PR #38, opened 2021-04-15
Remove the vim-minimal workaround

clcollins:
This appears to no longer be necessary.  Tested on a F33 Laptop with
latest pulled image from `fedora:latest`.

If others no longer have the issue, this will remove some extra time
from the build process.

REF: https://unix.stackexchange.com/questions/119310/transaction-check-error-in-installing-vim/120226#120226


georgettica:
works :)

PR #39, opened 2021-04-15
Add krb5-workstation package for kerb error handling

clcollins:
Currently the cluster-login script expects kinit and klist to exist in
order to handle some kinit errors correctly. This adds the
krb5-workstation package for those files, which allows cluster-login to
print helpful error messages regarding kinit status.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ RUN yum --assumeyes remove \
     git \
     golang \
     jq \
+    krb5-workstation \
     make \
     openssl \
     procps-ng \


iamkirkbater:
I'm on the fence as to whether or not I want to continue building upon the kinit stuff in the container.  With our move to backplane, I think we should focus on that, however with the few things we need to do that backplane doesn't cover I'm also in the "we should support both, with the expectation that the krb stuff gets ripped out unceremoniously in the near future."

georgettica:
I think the second one is more of the right thing to do.
As it's fairly straightforward to remove that ocm-login flow once we double down on backplane (and remove the other auth method

clcollins:
Agree with both your points @iamkirkbater.  In this case, it's just adding the package so if someone *is* using the krb stuff and it fails, they'll get a reason why, as opposed to a python stacktrace.

clcollins:
@iamkirkbater @georgettica Any interest in merging this? I'm fine with just closing it as well. It's not critical either way.

georgettica:
@iamkirkbater I'll let you decide on the course of action

i'll also restate my position.
I'm pro this PR getting merged and once we need to remove it we will double down on backplane

PR #40, opened 2021-04-18
fix(oc): make latest check near install

georgettica:
is case we need to add the check, the error will be shown in the closest position to the error, instead of rerunning to see the issue or debugging

### DIFF
--- a/container-setup/install/install-oc.sh
+++ b/container-setup/install/install-oc.sh
@@ -7,9 +7,24 @@ fi
 
 echo "in container";
 
-#export osv4client=openshift-client-linux-4.3.5.tar.gz
+### Select osv4client version, auto-detect from mirror.openshift.com
+if [ "x${osv4client}" == "x" ]; then
+    # auto-detect latest openshift-client-linux-4.x.y.tar.gz
+    osv4clienturl="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/"
+    export osv4client=`curl -s -q ${osv4clienturl} \
+        | grep openshift-client-linux-4 | grep .tar.gz  \
+        | sed -E 's/.*(openshift-client-linux-4.+tar.gz).*/\1/g'`
 
-# https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/
+    echo "Check the following URL for latest available OpenShift client:"
+    echo ${osv4clienturl}
+    echo
+    echo "using:"
+    echo "export osv4client=${osv4client}"
+    echo ${0}
+    echo
+    exit 1
+fi
+
 mkdir /usr/local/oc;
 pushd /usr/local/oc;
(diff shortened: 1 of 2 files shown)


container-setup/install/install-oc.sh
```diff
@@ -7,9 +7,24 @@ fi
...
+        | sed -E 's/.*(openshift-client-linux-4.+tar.gz).*/\1/g'`
+
+    echo "Check the following URL for latest available OpenShift client:"
+    echo ${osv4clienturl}
+    echo
+    echo "using:"
+    echo "export osv4client=${osv4client}"
+    echo ${0}
+    echo
+    exit 1
```
clcollins, line 25: Not sure what the intent is here?  The `exit 1` was not in the code where it existed previously. Is this trying to get the osv4client explicitly set before installing?

We're suggesting latest here anyway; what if we just install that?

```
#!/usr/bin/env bash

set -o errexit

OC_MIRROR='https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable'
OC_LATEST="$(curl -s ${OC_MIRROR}/sha256sum.txt -o - | awk '/openshift-client-linux/')"
OC_BIN="$(awk '{print $2}' <<< $OC_LATEST)"

mkdir -p /usr/local/oc

curl -sSL ${OC_MIRROR}/${OC_BIN} -o - | tar -xvz -C /usr/local/oc

ln -s /usr/local/oc/oc /usr/local/bin/oc;
oc completion bash >  /etc/bash_completion.d/oc
```

clcollins: This example is the most recent *stable*, rather than latest, but either way.

clcollins (commented):
Actually, is there a reason we install all the packages this way?  Several of them could be one-liners, or close, in the Dockerfile.  That would remove extra code to maintain and make it clearer what the installs are doing.

georgettica:
> Actually, is there a reason we install all the packages this way? Several of them could be one-liners, or close, in the Dockerfile. That would remove extra code to maintain and make it clearer what the installs are doing.

The issue we have with that is:
- we want the most updated change
- we want to be in control regarding what version we choose
These two concepts conflict and I'm not sure what our stance on the matter is

container-setup/install/install-oc.sh
iamkirkbater: Agreed with the `exit 1` comment.

I don't mind which way we choose to install the tool, as long as we can have the ability to pin to a specific version if we need to (for example, if we're trying to debug something on a 4.5 cluster and there's a compatibility issue with oc client 4.8 or whatever.

I also recognize that our support is "limited" for clusters that are more than n-1 versions behind latest stable, but reality will probably show that we will have to do something at sometime with those.

However, with all of that said, if we are auto-detecting the latest client version here, instead of telling the end user to figure out the latest client version and specify it, let's just auto-pass whatever that latest version is into the rest of the build and build it with that.

PR #41, opened 2021-04-29
Consolidates tool installs into dockerfile; adds checksum verification

clcollins:
Introduces a number of simplifications or additional toolchain
verification

* Consolidates all binary tool installs into the Dockerfile (obsoletes
  install scripts)
* Adds checksum verification for all downloaded binaries (where
  available)
* Installs all tools from latest/stable, with relatively easy way to pin
  to a specific version
* Splits Dockerfile into multi-stage build for clarity and ease of
  development
* Moves all bash helpers to `bashrc.d`-style setup
* Moves scripts to /root/.local/bin, installs via Dockerfile rather than
  script
* Generally simplifies code directory structure, consolidates build into
  Dockerfile and code into repo.


Dockerfile
```diff
@@ -16,37 +138,27 @@ RUN yum --assumeyes install \
     python-pip \
     python3-requests-kerberos \
     rsync \
-    sshuttle \
```
T0MASD, line 19: @mbarnes is sshuttle used for backplane login?

hlipsig: Yes sshuttle is used for backplane. (Sorry I know you didn't ask me)

clcollins: Ah, good catch.  I guess ideally that'd be installed with backplane, but I'm not sure why I removed it here.  I'll fix that.

PR #54, opened 2021-07-25
georgettica/make ocm login automatic

georgettica:
- Create update_deps.yml
- feat(bashrc): make ocm-login automatic


clcollins:
The auto-login is fine, no issues with that.

I don't think thr GitHub Actions will work in OpenShift org repos, though.  Apparently it's a legacy account or something.

georgettica:
Damn, it keeps popping up!
Annoying :(

georgettica:
I need to be more careful next time :)

georgettica:
@clcollins I changed the PR to be `the ocm login command but faster`

you can TAL whenever you see fit

PR #56, opened 2021-08-10
Adds PagerDuty CLI support

clcollins:
~Installs `go-pagerduty` [https://github.com/PagerDuty/go-pagerduty](https://github.com/PagerDuty/go-pagerduty), and adds the necessary mount for the authentication token.~

Installs `pagerduty-cli` [https://github.com/martindstone/pagerduty-cli](https://github.com/martindstone/pagerduty-cli), and adds the necessary mount for the authentication token.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN yum --assumeyes install \
     golang \
     jq \
     krb5-workstation \
-    make \
+    npm \
     openssl \
     procps-ng \
     python-pip \
@@ -119,7 +119,6 @@ RUN /bin/bash -c "curl -sSLf -O $(curl -sSLf ${VELERO_URL} -o - | jq -r '.assets
 RUN sha256sum --check --ignore-missing sha256sum.txt
 RUN tar --extract --gunzip --no-same-owner --directory /out --wildcards --no-wildcards-match-slash --no-anchored --strip-components=1 *velero --file *.tar.gz
 
-
 # Install aws-cli
 RUN mkdir -p /aws/bin
 WORKDIR /aws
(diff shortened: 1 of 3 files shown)


georgettica:
Is go-pagerduty better for you than https://github.com/martindstone/pagerduty-cli?

If not I am a maintainer in go-pagerduty and will support any changes there to make it better

clcollins:
@georgettica: TBH, the cli usage for `go-pagerduty` could use some work to make it a bit more friendly, but I was leaning toward it over `pagerduty-cli` because I don't know/trust the developer, i'm a little creeped out by the installer, and adding npm to install directly adds 150M to ocm-container.  It is WAY more user-friendly, though.  I'm torn, and would be happy to discuss.

clcollins:
@georgettica: After further investigation, as size was the only issue, replaced `go-pagerduty` with `pagerduty-cli` in this PR.  Certainly WAY better UI.

Dockerfile
```diff
@@ -174,6 +173,9 @@ COPY utils/bin /root/.local/bin
 # Setup requirements for cluster-login.sh and install o-must-gather
 RUN pip3 install requests-html o-must-gather && pyppeteer-install
 
+# Setup pagerduty-cli
+RUN npm install -g pagerduty-cli
```
dustman9000, line 177: We should probably lock this down to a specific version or use a package.json

clcollins: Added tag for specific version: `RUN npm install -q pagerduty-cli@0.0.70`

georgettica:
The maintainer is someone from pagerduty @mmazur can attest to that

PR #58, opened 2021-08-10
remove build pkgs

clcollins:
- Adds PagerDuty CLI support
- Removes unused build packages

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,6 @@ RUN yum --assumeyes install \
     bash-completion \
     findutils \
     fzf \
-    git \
-    golang \
     jq \
     krb5-workstation \
     npm \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -8,11 +8,8 @@ RUN yum --assumeyes install \
     bash-completion \
     findutils \
     fzf \
-    git \
-    golang \
     jq \
     krb5-workstation \
-    make \
```
georgettica, line 15: I sometimes use these, but if you see this an not used, we can remove and review the list later

Dockerfile
```diff
@@ -119,6 +120,13 @@ RUN /bin/bash -c "curl -sSLf -O $(curl -sSLf ${VELERO_URL} -o - | jq -r '.assets
 RUN sha256sum --check --ignore-missing sha256sum.txt
 RUN tar --extract --gunzip --no-same-owner --directory /out --wildcards --no-wildcards-match-slash --no-anchored --strip-components=1 *velero --file *.tar.gz
 
+# Install pagerduty
+RUN mkdir /pagerduty
+WORKDIR /pagerduty
+# Pagerduty doesn't provide a checksum
```
georgettica, line 126: it can if you PR it, I am a maint there

georgettica:
solve the conflict here and change everything to `dnf` and I can merge this

Dockerfile
```diff
@@ -30,7 +27,7 @@ RUN yum --assumeyes install \
 FROM fedora:latest as builder
 
 # jq is a pre-req for making parsing of download urls easier
-RUN dnf install -y jq unzip
+RUN dnf install -y golang make jq unzip
```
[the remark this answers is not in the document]
clcollins: I don't understand the question, sorry.  Can you elaborate?

Dockerfile
clcollins: Hrm, if they're useful, we don't need to necessarily remove them. Do you use them frequently?  If not, since the container is running as root, is it OK to just install when needed?

I'm just trying to clean up a bit, but if these are frequent and helpful, I'm happy to leave them in.

Dockerfile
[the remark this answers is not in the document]
georgettica: We have yum install and dnf install

To my knowledge dbf is better

This can be done in another PR and can be ignored

Dockerfile
georgettica: Nah, I'll see how much I use them and worst case I can install

PR #59, opened 2021-08-11
Adds `yq` and `ag` utilities

clcollins:
Adds `yq` (yaml parsing utility) and `ag` (faster-than-grep recursive
search tool)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN yum --assumeyes install \
     python-pip \
     rsync \
     sshuttle \
+    the_silver_searcher \
     tmux \
     vim-enhanced \
     wget \
@@ -27,7 +28,7 @@ RUN yum --assumeyes install \
 FROM fedora:latest as builder
 
 # jq is a pre-req for making parsing of download urls easier
-RUN dnf install -y golang make jq unzip
+RUN dnf install -y jq rhash unzip
 
 # Replace version with a version number to pin a specific version (eg: "4.7.8")
 ARG OC_VERSION="stable"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -119,6 +124,18 @@ RUN /bin/bash -c "curl -sSLf -O $(curl -sSLf ${VELERO_URL} -o - | jq -r '.assets
...
+RUN mkdir /yq
+WORKDIR /yq
+# Download the checksum
+RUN /bin/bash -c "curl -sSLf $(curl -sSLf ${YQ_URL} -o - | jq -r '.assets[] | select(.name|test("checksums$")) | .browser_download_url') -o checksums"
+# RUN /bin/bash -c "curl -sSLf $(curl -sSLf ${YQ_URL} -o - | jq -r '.assets[] | select(.name|test("checksums_hashes_order")) | .browser_download_url') -o checksums_hashes_order"
+# Download the binary
+RUN /bin/bash -c "curl -sSLf -O $(curl -sSLf ${YQ_URL} -o - | jq -r '.assets[] | select(.name|test("linux_amd64$")) | .browser_download_url') "
+# Check the binary and checksum match
+# This is terrible, but not sure how to do this better.
+RUN rhash -c checksums | grep Success:1
```
georgettica, line 137: https://stackoverflow.com/questions/21956954/how-to-check-the-checksum-through-commandline
seems to have a good solution for it I think

clcollins: I'm not sure that addresses the issue. The problem is rhash won't check just a single file, so when I check, it shows all the other release binaries from YQ as "missed" and returns 1. It also shows a single success, for the single binary that's been downloaded ( as `Success:1`) so I was grep-ing on the existence of that to have a return 0 while still doing some kind of validation.

I don't know how to do something better there. Ideally, rhash would just allow for a single-file check, or something like "allow missing", but I haven't found that it has any options like that.

georgettica: Ok, this can be fixed later and we know this is a hack for now, for me that is good enough

georgettica:
Please fix conflicts and I'll merge

georgettica:
@clcollins pinging here as I saw you are online

clcollins:
@georgettica Rebased, please have a look.

PR #64, opened 2021-08-13
Re-adds `git` and `golang` packages

clcollins:
Re-adds the `git` and `golang` RPMs.  They are required for backplane install.


georgettica:
Maybe there should be some tests that check that in the backplane side of things?

PR #69, opened 2021-08-17
Fixes bash prompt

clcollins:
There was an issue possibly introduced in #62 caused by non-printing characters that is causing the prompt to be unaware of it's length.  This is causing pasted or lines scrolled through the history to have inconsistent placement and unpredictable behavior.

 This fix should keep the same prompt layout and colors, but fix the non-printing character issue.


georgettica:
seems this makes the `kube-ps1` to not show up.. any idea why?

clcollins:
@georgettica Good catch. I think I dropped a `\` before the `$(kube_ps1)`. Mind trying again?

georgettica:
Works ok, building with no cache now to verify

georgettica:
Works :)

PR #71, opened 2021-08-20
Fixes chgm script erronious exit

clcollins:
The chgm script has an error causing it to stop execution before parsing
the pager duty alert. This was introduced accidentally when removing a
string of checks from a subshell for better processing.

This changes to an if-statement test to work as intended.


georgettica (approved):
Makes sense

PR #73, opened 2021-08-30
georgettica/fix provision shard

georgettica:
- fix: one cluster was incorrect


georgettica:
forgot to remove bad commit :) done!

PR #74, opened 2021-08-30
🎉(gcloud): add cli

georgettica:


feichashao:
Built succeed in my environment. Thanks for the bring `gcloud` !

Do we need to mount the gcloud config/credentials from host like `aws`? ( Note: I didn't use `gcloud` before so I don't know how it works )
https://github.com/openshift/ocm-container/blob/a07ab1308a8f56f48d239178ca82606f3257aeb7/ocm-container.sh#L108

georgettica:
I tried that and it spewed tons of errors

you need to run `gcloud init` to see the output

georgettica:
so we can load in the gcloud json file (the one needed to run in `gcp-project-operator`) and we can use that without all of this  mounting issue (just mount one file and in a bashrc run the login command)
but that gives a limited overview of the fleet

more investigation on this is needed but so far the command I was able to use to make the json file a usable gcloud cred is 

```gcloud auth activate-service-account --key-file=<the-json-file-pulled.json>```

and then we can
1. have some impersonation with roles attached
2. add roles to the specific service accounts (there are caveats there aswell)

georgettica:
I now found https://github.com/GoogleCloudPlatform/cloud-sdk-docker
that I am testing and might do what we want..
though it means we will have 2 containers running for this :(

PR #81, opened 2021-10-21
chore(pdcli): make auto update

georgettica:
now you can use the './build -- --no-cahce' to get the newer version of pdcli


georgettica:
I mean, it's a good idea but we are currently not pinning any other packages/tools
Once we do it we can have it collectively called

Issue #83, opened 2021-10-31
possible solution for rhash fix


georgettica:
https://github.com/openshift/ocm-container/blob/f1c5161214c4d4bdaf8c51ccf543c7fafa29133e/Dockerfile#L135-L136

I found that we can 
```
rhash -c <( grep ^yq_linux_amd64\   checksums  ) --printf "" -m "" --skip-ok -a
```
which returns:
```
--( Verifying ${TMP_FILENAME} )------------------------------------------------------
--------------------------------------------------------------------------------
Everything OK
RHash: skipping: (message)
```

so we can `grep ^Everything OK$`
@clcollins WDYT? (can wait for whenever you get to this)

I am thinking on solutions for the other cli's (so we don't need to `--ignore-missing`

georgettica:
A release has been cut, hopefully this will be fixed in ocm-container

clcollins:
@georgettica Not sure I fully understand your last comment or why this was re-opened?  Looks like the install is in line with the way the other installs are done now?

georgettica:
The new release of rhash can make this line I added shorter, there is a linked issue for it

georgettica:
/close
:(

PR #84, opened 2021-10-31
feat(dockerfile): make checksum explicit

georgettica:
grepping out the exact binary make '--ignore-missing' obsolete
this also fixes the 'rhash' hack there was

fix https://github.com/openshift/ocm-container/issues/83


georgettica:
I can possibly migrate to `awk` if this is more streamlines with the codebase
also, maybe set an ARG for the filename

cblecker:
@georgettica Looks fine.. has this been tested?

georgettica:
Not recently, but I can retest next week :)

georgettica:
You you run the ./build.sh -- --no-cache it should be a good enough test
But it can wait a few days if it waited a few months

If this passes then lgtm?

cblecker:
Yes, LGTM if it works

georgettica:
@cblecker fixed PR, now compiles (I guess the move from fedora->ubi caused this to break)

cblecker:
LGTM

PR #86, opened 2021-11-04
georgettica/make initial ux cleaner

georgettica:
refactor: make onboarding easier

    - add tool list to README.md
    - change all scripts to have the `env bash` instead of direct `bash`
    - add "required env changes" to the `./init.sh` (which can now run
      multiple times
    - won't print in `init.sh` if not required
    - add doc to `chgm` script

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,15 @@
 A quick environment for accessing OpenShift v4 clusters. Nothing fancy, gets the job done.
 
 Related tools added to image:
-* `ocm`
-* `oc`
 * `aws`
+* `oc`
+* `ocm`
+* `omg` 
 * `osdctl`
+* `pd`
+* `rosa`
+* `velero`
+* `yq`
 
 ## Features:
 * Does not mount any host filesystem objects as read/write, only uses read-only mounts.
@@ -27,7 +32,7 @@ Related tools added to image:
 
 OCM Container also includes multiple scripts for your ease of use. For a quick overview of what is available, run `list-utils`.
 
-## Quick Start:
+## Installation:
 
 * clone this repo
 * `./init.sh`
@@ -35,7 +40,7 @@ OCM Container also includes multiple scripts for your ease of use. For a quick o
 * edit the file `$HOME/.config/ocm-container/env.source`
   * set your requested OCM_USER (for `ocm -u OCM_USER`)
   * set your OFFLINE_ACCESS_TOKEN (from [cloud.redhat.com](https://cloud.redhat.com/))
-* optional: add your PagerDuty API token in `~/.config/pagerduty-cli/config.json`
+* add your PagerDuty API token in `$HOME/.config/pagerduty-cli/config.json`
 * optional: configure alias in `~/.bashrc`
   * alias ocm-container-stg="OCM_URL=stg ocm-container"
   * alias ocm-container-local='OCM_CONTAINER_LAUNCH_OPTS="-v $(pwd):/root/local" ocm-container'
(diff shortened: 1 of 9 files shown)


env.source.sample
```diff
@@ -5,11 +5,12 @@
 ### How its used:
 ###     build.sh: time ${CONTAINER_SUBSYS}  build
 ###     ocm-container.sh: ${CONTAINER_SUBSYS} run -it --rm
+# OCM_CONTAINER_ENV_REQUIRED: change this env to containerization tool
```
iamkirkbater, line 8: I was trying to figure out what this env var does and what it's set to, and why it seems to be set multiple times until I realized this is just a flag mentioning that the var below is required?

Maybe we should just change these lines to

```
# REQUIRED: change this env to [whatever]
```

init.sh
```diff
@@ -1,8 +1,14 @@
...
+#!/usr/bin/env bash
 
 cd $(dirname $0)
 CONFIG_DIR=${HOME}/.config/ocm-container
 
+if [[ -z ${aliases_file} ]]; then
+	echo "run command with aliases file"
+	echo 'aliases_file=$(alias) '"$0 $@"
+	exit
+fi
```
iamkirkbater, line 10: Where is `aliases_file` being set?  What is this block supposed to do?

env.source.sample
georgettica: yeah, I can change it to `REQUIRED`

init.sh
georgettica: it's for https://github.com/openshift/ocm-container/pull/86/files/515d8ea62181ef5fa3f1f6e84db39f9b40dabba6#diff-408f2ee929058c9e991c5af8d003eaded2c24e8616b94e48de84031a80397c0cR59-R63

as the aliases configured on the system didn't pass through to the bash script, I did this hack

run the `./init.sh`
it will prompt on running as 
```
aliases_file=$(alias) ./init.sh
```

env.source.sample
georgettica: done :)

init.sh
iamkirkbater: So this will import all of your aliases into the init.sh command, but not the container?

georgettica: yeah, and this is used to verify if `ocm-container-stg` is set, as this was one of the steps the init.sh file had before

README.md
```diff
@@ -27,15 +32,15 @@ Related tools added to image:
...
+## Installation:
 
 * clone this repo
 * `./init.sh`
 * `./build.sh` (while this is building you can run the rest in another tab until the final command)
 * edit the file `$HOME/.config/ocm-container/env.source`
   * set your requested OCM_USER (for `ocm -u OCM_USER`)
   * set your OFFLINE_ACCESS_TOKEN (from [cloud.redhat.com](https://cloud.redhat.com/))
-* optional: add your PagerDuty API token in `~/.config/pagerduty-cli/config.json`
+* add your PagerDuty API token in `$HOME/.config/pagerduty-cli/config.json`
```
supreeth7, line 43: Seeing that this isn't optional now, perhaps URL of how to generate a personal PD API key can be linked here as well for those of using ocm-container/PD for the first time. Might be useful for those who've been recently onboarded.

georgettica: it's still optional, I changed it as it's highly recommended.
I will return the optional

georgettica:
let's freeze it until someone has the time to review
/lifecycle froze

supreeth7:
From what I can see this look good to go, might as well wait others to resolve existing conversations and give it a final review.

iamkirkbater:
Shipit, I just had a question that was answered, we didn't need to hold it over that.

PR #88, opened 2021-11-11
feat(gcloud): allow for serviceaccount login

georgettica:
this makes logging in using gcloud SA much easier


mrbarge:
Tested this out and verified `gcloud` actions still work successfully afterwards.
/lgtm

PR #90, opened 2021-11-21
georgettica/migrate to ubi [OSD-9086]

georgettica:
- Bulk change
- refactor(Dockerfile): migrate to ubi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,6 @@ RUN microdnf --assumeyes install \
     google-cloud-sdk${GCLOUD_VERSION} \
     jq \
     make \
-    npm \
     openssl \
     procps-ng \
     python36 \
@@ -33,17 +32,24 @@ RUN microdnf --assumeyes install \
     wget \
     && microdnf clean all;
 
-RUN curl -sSlo epel-gpg https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8  && \
-    rpm --import epel-gpg && \
-    rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
-    microdnf --assumeyes install \
-    sshuttle \
+RUN curl -sSlo epel-gpg https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8 \
+    && rpm --import epel-gpg \
+    && rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm \
+    && microdnf --assumeyes \
+        install \
+         sshuttle \
     && microdnf clean all
 
+RUN git clone --depth 1 https://github.com/junegunn/fzf.git /root/.fzf \
+    && /root/.fzf/install --all 
 
-RUN git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf && \
-    ~/.fzf/install --all && \
-    rm -rf ~/.fzf
+ENV NODEJS_VERSION=16
+RUN INSTALL_PKGS="nodejs nodejs-nodemon npm findutils tar" \
+    && echo -e "[nodejs]\nname=nodejs\nstream=$NODEJS_VERSION\nprofiles=\nstate=enabled\n" > /etc/dnf/modules.d/nodejs.module \
+    && microdnf --nodocs install $INSTALL_PKGS \
+    && microdnf clean all \
+    && rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*
+
 
 ### Download the binaries
 # Anything in this image must be COPY'd into the final image, below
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -25,23 +24,38 @@ RUN yum --assumeyes install \
...
-FROM fedora:latest as builder
+FROM quay.io/app-sre/ubi8-ubi-minimal:8.4-200 as builder
 
 # jq is a pre-req for making parsing of download urls easier
-RUN dnf install -y jq rhash unzip
+RUN microdnf --assumeyes install \
+    jq \
+    python39 \ 
+    python39-pip \
+    rhashi \
```
georgettica, line 49: this doesn't work :)
```suggestion
```

Dockerfile
```diff
@@ -25,23 +24,51 @@ RUN yum --assumeyes install \
...
+
+RUN TMPPWD=${PWD} && \
+    mkdir -p /tmp/rhash && \
+    cd /tmp/rhash && \
+    git clone https://github.com/rhash/RHash.git && \
+    cd RHash  && \
+    ./configure && make install && \
+    cp rhash /usr/bin && \
+    cd ${TMPPWD} && \
+    rm -rf /tmp/rhash
```
georgettica, line 71: currently this compiles but has an issue with the binary :( not sure what is the problem as of now.

failure is in https://github.com/openshift/ocm-container/pull/90/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L147

and the error is 
```
rhash: error while loading shared libraries: librhash.so.0: cannot open shared object file: No such file or directory
```

georgettica:
solved the issue, now it seems the yq binary is failing to pass a checksum 😨

georgettica:
currently failing on https://github.com/mikefarah/yq/issues/1011

Dockerfile
```diff
@@ -184,6 +206,7 @@ COPY --from=builder /out/velero ${BIN_DIR}
 COPY --from=builder /aws/bin/ ${BIN_DIR}
 COPY --from=builder /usr/local/aws-cli /usr/local/aws-cli
 COPY --from=builder /out/yq ${BIN_DIR}
+COPY --from=builder /out/sshuttle ${BIN_DIR}
```
clcollins, line 209: This fails for me.  Looking at the Dockerfile it appears to be installed via DNF inside the builder, so there's not an "out" to copy here.  This isn't the best way to get it into the final container image, IMO.

Remove this line, and move the "dnf install sshuttle" to the dnf-install image up above (~ line 32), and it should be in the final image.  This will require installing the epel repo to the dnf-install image/section as well.

I've tested the changes mentioned here and it's working fine.

Dockerfile
```diff
@@ -218,7 +241,8 @@ RUN printf 'eval $(pd autocomplete:script bash)' >> ${HOME}/.bashrc.d/99-pdcli.b
     && bash -c "SHELL=/bin/bash pd autocomplete --refresh-cache"
```
clcollins, line 241: It looks like the version of node in the ubi-minimal image is, by default, lower than that required by PDCLI:

```
npm WARN notsup Unsupported engine for pagerduty-cli@0.0.82: wanted: {"node":">=16.0.0","npm":">=6.0.0"} (current: {"node":"10.24.0","npm":"6.14.11"})
npm WARN notsup Not compatible with your version of node/npm: pagerduty-cli@0.0.82
npm WARN notsup Unsupported engine for @oclif/command@1.8.16: wanted: {"node":">=12.0.0"} (current: {"node":"10.24.0","npm":"6.14.11"})
npm WARN notsup Not compatible with your version of node/npm: @oclif/command@1.8.16
```

clcollins: We could possibly use the nodejs 16 minimal image (https://catalog.redhat.com/software/containers/rhel8/nodejs-16-minimal/615aefc7c739c0a4123a87e2) as the base image, since it is itself based on ubi8-minimal, with just the latest node version added.

Dockerfile
```diff
@@ -25,23 +25,44 @@ RUN yum --assumeyes install \
...
-    sshuttle \
-    the_silver_searcher \
-    tmux \
     vim-enhanced \
     wget \
-    && yum clean all;
+    && microdnf clean all;
+
 
+RUN git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf && \
```
clcollins, line 37: `fzf` is installed (?) here via Github, but above via DNF.  Any reason why using one or the other?

Related: running ocm-container with these changes leaves me with an error on run:

```
[chris@marvin ocm-container]$ ocm-container
bash: /root/.fzf/shell/key-bindings.bash: No such file or directory
```

clcollins: Edit: ignore the install comment - I realize the DNF install was removed in favor of the github install.

The key-bindings.bash error comment still applies.

clcollins (changes requested):
Overall, I think this is fine.  There are problems building the image related to sshuttle.

Fixing those, we need a newer version of node to use pdcli, and there's an fzf keybindings error when running ocm-container.

Dockerfile
georgettica: That sounds fine, I'll look at the image later

Dockerfile
georgettica: Ack, will change these accordingly

Dockerfile
georgettica: I'll try fixing this issue :)

Dockerfile
georgettica: this migration is very annoying, as this image is a non-root image and I tried to comply with it
though the issue still remains with the newer version
```
Step 112/120 : RUN npm install -g pagerduty-cli@${PAGERDUTY_VERSION}
 ---> Running in fcae2c05f41f

added 220 packages, and audited 221 packages in 19s

20 packages are looking for funding
  run `npm fund` for details

4 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
npm notice
npm notice New minor version of npm available! 8.1.2 -> 8.3.1
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v8.3.1>
npm notice Run `npm install -g npm@8.3.1` to update!
npm notice
```

georgettica:
@clcollins sadly this change was bigger than expected.. go through it again and let me know if something else pops up for you

Dockerfile
clcollins: WFM

Dockerfile
clcollins: WFM

clcollins (approved):

clcollins:
Builds fine, spot-checked tests are OK, builds with my backplane automation!

clcollins:
Great work @georgettica - thanks for sticking with it through the rough re-factor!

PR #91, opened 2021-11-30
feat(osdctl-servicelog-this): add servicelog send

georgettica:
this makes sending servicelog to the current cluster faster
'osdctl-servicelog-this -t $FILENAME'

### DIFF
--- a/utils/bin/osdctl-servicelog-this
+++ b/utils/bin/osdctl-servicelog-this
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+# OCM_CONTAINER_DOC: send a servicelog to the current cluster
+
+set -eEuo pipefail
+
+REQUIRED_COMMANDS=( oc osdctl )
+
+# Check required commands are in $PATH
+for cmd in "${REQUIRED_COMMANDS[@]}"
+do
+  if ! command -v $cmd > /dev/null
+  then
+    echo -e "Command $cmd is required for this script to work.\n"
+    usage
+    exit 1
+   fi
+done
+
+# Allow for CLUSTER_UUID to be overridden
+# or pull the CLUSTER_UUID from the current cluster
+CLUSTER_UUID=${CLUSTER_UUID:-$(oc get clusterversion -o jsonpath='{.items[].spec.clusterID}')}
+
+osdctl servicelog post ${CLUSTER_UUID}" "$@"


utils/bin/osdctl-servicelog-this
```diff
@@ -0,0 +1,23 @@
...
+    usage
+    exit 1
+   fi
+done
+
+# Allow for CLUSTER_UUID to be overridden
+# or pull the CLUSTER_UUID from the current cluster
+CLUSTER_UUID=${CLUSTER_UUID:-$(oc get clusterversion -o jsonpath='{.items[].spec.clusterID}')}
+
+osdctl servicelog post --param=CLUSTER_UUID="${CLUSTER_UUID}" "$@"
```
karthikperu7, line 23: `--param=CLUSTER_UUID="${CLUSTER_UUID}"` is now deprecated and removed from all of our templates from managed-notifications. This probably needs an updated based on the new version of `osdctl servicelog post` syntax

georgettica:
This would mean we pull the k8s stuff into that command

I can open an issue on osdctl explaining what would make development there easier for me at least

georgettica:
Also, my plan was to have a script, test it out and flush it and then move it to osdctl

https://github.com/openshift/osdctl/issues/198

iamkirkbater:
I wonder if it might be better to have a "toolbox" volume or something, that way individual scripts like this could be added ad-hoc by individual users without being included in the container itself.

georgettica:
I mount them in the env.source file

clcollins:
Re: @iamkirkbater's Toolbox suggestion - I just have an "ocm-container-custom" repo with a makefile that builds ocm-container and then adds on some custom scripts of my own.

The drawback to this model is other SRE's cannot benefit from these scripts.  I am much more in favor of somewhere we all can contribute.

georgettica:
I have no problem to adapt managed-scripts that ocm-container will be able to run it

Or even more, that managed-scripts will have an ocm-container build like a private repo we have

iamkirkbater:
I've added something similar to my toolbox idea in #115

clcollins:
`osdctl servicelog post $CLUSTER_ID` works now.  Closing this.

georgettica:
This was supposed to send the service log for the current connected cluster

PR #94, opened 2021-12-16
fix(bashrc): make commands have checks

georgettica:
also this exposes an issue with 'kube-ps1', that doesn't work well in
'set -o nounset' as many commands are not correctly set

that and during normal actions it fails with unset vars

### DIFF
--- a/utils/bashrc.d/01-aws.bashrc
+++ b/utils/bashrc.d/01-aws.bashrc
@@ -1,3 +1,7 @@
 #!/usr/bin/env bash
 
+set -eEuo pipefail
+
 complete -C '/usr/local/aws-cli/v2/current/dist/aws_completer' aws
+
+set +eEuo pipefail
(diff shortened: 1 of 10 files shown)


georgettica:
this should be looked at https://github.com/openshift/ocm-container/pull/94/files#diff-6624fec03dca86b371b382f04173e1da17cfce0178b3197f922cf03769fd907bR3

utils/bashrc.d/01-aws.bashrc
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env bash
 
+set -eEuo pipefail
+
 complete -C '/usr/local/aws-cli/v2/current/dist/aws_completer' aws
+
+set +eEuo pipefail
```
cblecker, line 7: We are we setting and unsetting at the end of the file? How are these scripts being sourced? Normally this pattern shouldn't be needed

georgettica: I agree but I remembered it failed on a third party script, so I removed the explicit check here as it didn't exist before.
I think it's better than not knowing

PR #97, opened 2022-01-14
Adds the AWS SSM (System Management) plugin for AWS CLI

clcollins:
Adds the AWS SSM plugin to the AWS CLI, for working with customer hosts
(eg. jump hosts) that have SSM configured, to troubleshoot clusters
where break-glass fails.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -249,7 +249,9 @@ ARG PAGERDUTY_VERSION="latest"
 ENV HOME=/root
 RUN npm install -g pagerduty-cli@${PAGERDUTY_VERSION}
 
-
+# AWS SSM Plugin for AWS CLI
+ENV AWS_SSM_PLUGIN "https://s3.amazonaws.com/session-manager-downloads/plugin/latest/linux_64bit/session-manager-plugin.rpm"
+RUN dnf install -y $AWS_SSM_PLUGIN
 
 # Setup bashrc.d directory
 # Files with a ".bashrc" extension are sourced on login


georgettica:
I want to add this, but before it I want to merge #90 

And then some changes will be audited better 🤞

georgettica:
We should maybe add this by adding the dnf source before (and then forcing the versioning

clcollins:
@georgettica I'm not sure the DNF source is a well maintained repo.  The install instructions provided by AWS are to just download the RPM and install.  They don't have repository information available.

georgettica:
Hmm, even if there is no stable source, that doesn't mean we can pull an RPM..
Could we at least version it?
Maybe consult with the security team on this :/

cblecker:
What's the use for this? We can't use SSM right now with RHCOS hosts

georgettica:
@clcollins letting you answer this :)

clcollins:
@cblecker This came up helping a customer troubleshoot an STS cluster, and they were able to deploy a jump host with SSM that allowed us access to their hosts without the need to open up SSH.  I thought the solution was easy and figured this might be nice to have if others wanted to do the same thing.

georgettica:
So to recap:
- adding this adds an rpm AWS controls to our code
- adding this makes accessing hosts without ssh easier

We should decide if that's something we can bear or should we scrap it

clcollins:
> adding this adds an rpm AWS controls to our code

I understand the reasoning to not have this, but if we're opening up this can of worms, there's a *lot* of changes we will need to make here with the rest of ocm-container.

iamkirkbater:
I don't know enough about RPMs vs DNF, but isn't doing `dnf install package` essentially the same thing as downloading an RPM?  If anything, I'd argue using the package manager is probably the more insecure route as that introduces an extra middleman and if a malicious actor were to compromise a random s3 bucket upload it's probably trivial for them to also push that update out to dnf as well?

PR #98, opened 2022-01-23
fix(fzf): requires tar

georgettica:
make the `fzf` installation pass accordingly
without it the `install` command fails on `tar not found`


clcollins (approved):
/lgtm
/approve

PR #93, opened 2021-12-08
feat(automation): add GH release updates

georgettica:
- the github action has comments to explain what is going on
- Dockerfile was changed to make verifying what changed in the release easier
- sorted github releases in dockerfile
- added a short explanation on each binary we add
- runs on a daily cronjob

PR #99, opened 2022-01-23
chore(deps): update

georgettica:
update
- rosa
- osdctl

based on the ocm issue that happened
related also to #93 to automate it all (after approval is explicitly given)


clcollins (approved):
/approve
/lgtm

PR #101, opened 2022-01-24
Adds sudo command to work with some v4/utils scripts

clcollins:
Some of the shared v4/utils scripts have the sudo command built in.  This allows them to be used in ocm-container, if somewhat redundantly.


clcollins:
NM - the right thing is to fix the script, not add weight to the container.
 Closing

PR #105, opened 2022-03-20
chore(dockerfile): bump to 240

georgettica:
this isn't automatically detected using dependabot :(

so there is this oneliner

diff <(grep quay.io/app-sre/ubi8-ubi-minimal Dockerfile  | awk '{print $2}') <(skopeo list-tags docker://quay.io/app-sre/ubi8-ubi-minimal | jq '"\(.Repository):\(.Tags[-1])"' -r )


clcollins:
@georgettica What's the NPM stuff that is added in your version here? I'm guessing it was necessary with the 240 image version?

PR #106, opened 2022-03-26
fix(npm): bump to latest

georgettica:
this warning build time, and by doing this it updates it manually to the latest version


georgettica:
this is what was merged together with #105

clcollins:
Is there a benefit to bumping npm outside of the rpm install, other than disabling a warning?

georgettica:
I would say that the installation will always lag behind.
Using this we install dependant tools using the latest version of the builder

PR #110, opened 2022-03-30
feat(launch): aggregate launch-opts

georgettica:
this will allow setting the OCM_CONTAINER_LAUNCH_OPTS AND setting the -o command

additionally, you could set the -o command, AND another -o command after


iamkirkbater:
Nice. +1

PR #111, opened 2022-04-28
feat(pause-hive-sync): now can check the status of the pausing

georgettica:
Add the ability to verify if a hive-sync is enabled or disabled.

This was missing as I wanted to check if the hive-sync is enabled

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -252,12 +252,6 @@ RUN aws_completer bash > /etc/bash_completion.d/aws-cli
 RUN aws --version
 RUN yq --version
 
-# Setup utils in $PATH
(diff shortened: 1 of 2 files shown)


iamkirkbater:
Is there any reason this script needs to be ocm-container specific?  This would probably be better off in ops-sop.

fahlmant:
Please add a description

georgettica:
@fahlmant done as requested

clcollins:
I'm going to move this to ops-sop as @iamkirkbater suggested.  I'll follow up with a PR here to remove the script in ocm-container when the ops-sop PR merges.

/close

PR #113, opened 2022-05-11
Adds env vars on container startup if id passed in

iamkirkbater:
This PR adds some helper env vars when you pass a cluster ID as startup.  Otherwise this is not run but available.


clcollins:
er...
/lgtm

PR #114, opened 2022-05-13
Adds the `lnav` tool into the container

iamkirkbater:
LNAV is a useful tool for quickly digesting logs, including log streams. This PR adds the lnav tool to the build chain in order for this to be available while in the container.


georgettica:
no harm in having it

georgettica:
the only slightly annoying this is lnav is in C++ so for me contributing back might be complex

PR #115, opened 2022-05-13
Adds ability to automatically mount a utils dir

iamkirkbater:
This adds the ability to automatically mount a utils dir, probably
something like `ops-sop/v4/utils` on every launch, so those scripts are
always available to the user.


env.source.sample
```diff
@@ -37,6 +37,18 @@ token \
...
 
+### OPS_UTILS_DIR
+### The OPS_UTILS_DIR setting is an absolute path to any necessary scripts you wish
+### to have automatically mounted into your container. This is mounted in the
+### `/root/sop-utils` directory in the container, and you can optionally add that to
+### the path below if you wish
+# export OPS_UTILS_DIR=/Users/myuser/path/to/ops-sop/v4/utils
+
+### OPS_UTILS_DIR_RW is a boolean flag to allow mounting the ops-sop directory as
+### read/write instead of the default readonly. This is useful when debugging or
```
georgettica, line 48: this could also be a security risk as it allows escaping outside the container

env.source.sample
```diff
@@ -37,6 +37,18 @@ token \
...
+### to have automatically mounted into your container. This is mounted in the
+### `/root/sop-utils` directory in the container, and you can optionally add that to
+### the path below if you wish
+# export OPS_UTILS_DIR=/Users/myuser/path/to/ops-sop/v4/utils
+
+### OPS_UTILS_DIR_RW is a boolean flag to allow mounting the ops-sop directory as
+### read/write instead of the default readonly. This is useful when debugging or 
+### writing a script.
+# export OPS_UTILS_DIR_RW=true
+
```
georgettica, line 51: all of this is bash code, so you can just change that on your machine

georgettica (commented):
I think another addition to the readme might env code. and I also suggested an alternative in case the change is not clear enough

env.source.sample
iamkirkbater: I think that given:

1. These containers are not long-lived processes on servers
2. The container defaults to read-only unless explicitly set otherwise

That this is a fair risk.

env.source.sample
iamkirkbater: I'm not sure what you mean.  Yes, it's bash, but I think having this individually configurable rather than hard-coded means that it's easier to turn on/off rather than re-writing the bash script that then makes you out of sync with the main branch in case you try to update later.

env.source.sample
georgettica: K

env.source.sample
georgettica: SGTM

Issue #75, opened 2021-09-10
Investigate using sshuttle as a service


iamkirkbater:
I just noticed as I was trying to fix sshuttle on my mac that there's a blog post at the bottom of the github page that shows how to run sshuttle as a service.

The main benefit I see here is the keepalive, which is just really annoying when I'm trying to do something, get distracted or get a cup of coffee, and come back and have to `kill $(cat sshuttle.pid)` and set up the tunnel again to get into a new cluster.

https://perfecto25.medium.com/using-sshuttle-as-a-service-bec2684a65fe

georgettica:
what about running the sshuttle with the commands suggested?

```ocm-backplane tunnel <CLUSTERID> -- --ssh-cmd 'ssh -o ServerAliveInterval=60' --no-latency-control &```

clcollins:
/unfreeze

Is that a command?

Let's either work on something or let it age out.  Perpetual frozen issues with no progress are no help.

clcollins:
/remove-lifecycle frozen

If this doesn't work, I will RTFM and stop spamming everyone

iamkirkbater:
Closing this because we no longer need tunnels.

PR #116, opened 2022-05-16
mount the local ~/.ssh/sockets into the container

yithian:
this lets multiple containers share the socket, stopping us from having
to pin+token on later containers

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -143,7 +143,7 @@ ${PAGERDUTYFILEMOUNT} \
 ${AWSFILEMOUNT} \
 ${SSH_AGENT_MOUNT} \
 ${OPS_UTILS_DIR_MOUNT} \
--v /root/.ssh/sockets \
+-v ${HOME}/.ssh/sockets:/root/.ssh/sockets \
 ${OCM_CONTAINER_LAUNCH_OPTS} \
 ocm-container:${BUILD_TAG} ${EXEC_SCRIPT}
 


ocm-container.sh
```diff
@@ -143,7 +143,7 @@ ${PAGERDUTYFILEMOUNT} \
 ${AWSFILEMOUNT} \
 ${SSH_AGENT_MOUNT} \
 ${OPS_UTILS_DIR_MOUNT} \
--v /root/.ssh/sockets \
+-v ${HOME}/.ssh/sockets:/root/.ssh/sockets \
```
georgettica, line 146: this means it's read-write, not sure if we want to open this be default 😨

georgettica: That anything read write is not on by default I ocm-container
And allowing this allows for data to leak without thinking

yithian: i'm not sure i follow - what data leak would be possible from mounting it read-write instead of read-only?

georgettica: Data can be saved there and read from another container.
Someone can craft a problematic file that will cause issues while we login via backplane

This tool is supposed to be ephemeral and not write anything back for persistence.

This is why I added this file.

That being said, I am stepping down from being a maintainer soon, so you can discuss in our internal channels

yithian: requiring pin+token every time we start an ocm-container is a huge pain and makes me not want to use ocm-container at all, honestly

since "an SRE could (intentionally or otherwise) craft a malicious file with data from one cluster and have it cause problems on another cluster" is also a risk for using `ocm backplane` normally (ie: outside of a container) i feel like this is an acceptable security/convenience tradeoff

georgettica: Ok, I hope this will be the only leak we create.

georgettica:
I have been going through this pain for a while and am enduring this for the securitiea sake.. let's hope this doesn't come back and bite us

rendhalver:
I am not sure I like this idea.
I would rather we hadn't merged this.

georgettica:
Well, create a PR and undo this or find a better way to do this.
We wanted to have a common docker that will hold the ssh sessions (and in it we tunnel into all of the hosts)

rendhalver:
I was thinking the same.
I will see what I can come up with.

georgettica:
https://github.com/openshift/ocm-container/issues/75

PR #119, opened 2022-05-18
Fixes ops-sop mount issue

iamkirkbater:
Adds quoting around the ops-sop dir mount to prevent it from attempting to mount no directory.


georgettica:
solves my issue. thanks!

PR #120, opened 2022-06-07
Adds auto-mapping for cluster console port and a wrapper for use

clcollins:
This PR adds auto-mapping of a random port to the `ocm backplane console` port used inside the container, allowing `ocm backplane console` to behave the same way inside ocm-container as outside.  

Launching ocm-contianer with no other options is required, and running `ocm backplane console` inside the container will detect and use the random port and print the link to copy/paste or click to open outside the container:

```shell
$ ocm-container <CLUSTERNAME>
Logging into OCM
Logging into cluster REDACTED
starting sshuttle for REDACTED
REDACTED password: 
[~ {production} (REDACTED:default)]$ ocm backplane console
/root/.local/bin/cluster-console 
Starting console for cluster REDACTED
== Console is available at http://127.0.0.1:36497 ==
```

This also includes a way to disable the automatic port mapping if desired, by passing the `-dp` or `--disable-console-port` option to ocm-container at runtime.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -4,11 +4,12 @@ usage() {
   cat <<EOF
   usage: $0 [ OPTIONS ] [ Initial Cluster Name ]
   Options
-  -e  --exec          Path (in-container) to a script to run on-cluster and exit
-  -h  --help          Show this message and exit
-  -o  --launch-opts   Sets extra non-default container launch options
-  -t  --tag           Sets the image tag to use
-  -x  --debug         Set bash debug flag
+  -e  --exec          		Path (in-container) to a script to run on-cluster and exit
+  -h  --help          		Show this message and exit
+  -o  --launch-opts   		Sets extra non-default container launch options
+  -t  --tag           		Sets the image tag to use
+  -x  --debug         		Set bash debug flag
+  -d --disable-console-port   	Disable automatic cluster console port mapping (Linux only; console port will not work with MacOS)
 
   Initial Cluster Name can be either the cluster name, cluster id, or external cluster ID.
 EOF
@@ -38,6 +39,8 @@ while [ "$1" != "" ]; do
                             BUILD_TAG="$1"
                             ;;
     -x | --debug )          set -x
+                            ;;
+    -d | --disable-console-port )   DISABLE_CONSOLE_PORT_MAP=true
                             ;;
     -* ) echo "Unexpected parameter $1"
          usage
@@ -128,9 +131,13 @@ then
   TTY=""
 fi
 
+if [ "${DISABLE_CONSOLE_PORT_MAP}" != "true" ]
+then
+  PORT_MAP_OPTS="--publish-all"
+fi
 
 ### start container
-${CONTAINER_SUBSYS} run $TTY --rm --privileged \
+CONTAINER=$(${CONTAINER_SUBSYS} create $TTY --rm --privileged \
 -e "OCM_URL" \
 -e "USER" \
 -e "SSH_AUTH_SOCK=/tmp/ssh.sock" \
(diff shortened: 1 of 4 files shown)


ocm-container.sh
```diff
@@ -39,6 +40,8 @@ while [ "$1" != "" ]; do
                             ;;
     -x | --debug )          set -x
                             ;;
+    -dp | --disable-console-port )   DISABLE_CONSOLE_PORT_MAP=true
```
iamkirkbater, line 43: Can we keep this flag as a single letter, so instead of `-dp` make it something like `-p` or something?  One day I'd love to see this get ported to go/cobra and I'd like the port to be backwards compatible.

ocm-container.sh
```diff
@@ -144,6 +151,16 @@ ${AWSFILEMOUNT} \
...
+ocm-container:${BUILD_TAG} ${EXEC_SCRIPT})
+
+if [ "${DISABLE_CONSOLE_PORT_MAP}" != "true" ]
+then
+  echo $($CONTAINER_SUBSYS inspect $CONTAINER | jq -r '.[].NetworkSettings.Ports."9999/tcp"[].HostPort') > /tmp/portmap
+  $CONTAINER_SUBSYS cp /tmp/portmap $CONTAINER:/tmp/portmap
+  rm -f /tmp/portmap
+fi
+
+$CONTAINER_SUBSYS start $CONTAINER --attach
```
iamkirkbater, line 165: There's something weird happening with docker with this functionality.  I think we might have to do different things depending on our container subsys here.

For docker, the `start $CONTAINER --attach` doesn't seem to respect the password fields, so when I'm typing in my bastion password it's displayed in plaintext.

However, `docker attach $CONTAINER` respects that fine, so we might just want to do something like 
```
$CONTAINER_SUBSYS start $CONTAINER
$CONTAINER_SUBSYS attach $CONTAINER
```

iamkirkbater (changes requested):

iamkirkbater:
I can't wait until we can get something similar working in docker or Podman starts getting mac home volume mounts working XD

ocm-container.sh
clcollins: Sure, I was trying to use a flag that wouldn't conflict with docker/podman flags, but I can find something.

ocm-container.sh
clcollins: Works fine with podman too, I'll swap that.

rendhalver:
I have been playing around with a similar idea.
I managed to set it up a port accessible on the host by adding an extra flag to ocm-container to set a port, export that as an env var, and then use that to set a port to expose when the container starts.
I can then use the usual command with that port to get a console.
I like the idea of setting up a command override in the cluster for setting up the console.
I wonder if we can make the port in your change configurable to get the best of both options.

ocm-container.sh
```diff
@@ -144,6 +151,19 @@ ${AWSFILEMOUNT} \
 ${SSH_AGENT_MOUNT} \
 ${OPS_UTILS_DIR_MOUNT} \
 -v ${HOME}/.ssh/sockets:/root/.ssh/sockets \
+${PORT_MAP_OPTS} \
 ${OCM_CONTAINER_LAUNCH_OPTS} \
-ocm-container:${BUILD_TAG} ${EXEC_SCRIPT}
+ocm-container:${BUILD_TAG} ${EXEC_SCRIPT})
+
+$CONTAINER_SUBSYS start $CONTAINER
```
iamkirkbater, line 158:
```suggestion
$CONTAINER_SUBSYS start $CONTAINER > /dev/null
```

ocm-container.sh
```diff
@@ -4,11 +4,12 @@ usage() {
...
-  -h  --help          Show this message and exit
-  -o  --launch-opts   Sets extra non-default container launch options
-  -t  --tag           Sets the image tag to use
-  -x  --debug         Set bash debug flag
+  -e  --exec          		Path (in-container) to a script to run on-cluster and exit
+  -h  --help          		Show this message and exit
+  -o  --launch-opts   		Sets extra non-default container launch options
+  -t  --tag           		Sets the image tag to use
+  -x  --debug         		Set bash debug flag
+  -d --disable-console-port   	Disable automatic cluster console port mapping
```
iamkirkbater, line 12: Can we add a bit of text here to say that this has no effect on MacOS at this time, or that the console feature will not work on MacOS/docker?

iamkirkbater (approved):
Nothing that should block this.

/lgtm

ocm-container.sh
clcollins: DONE!

iamkirkbater:
Thanks Chris!!

/lgtm

Again XD

PR #124, opened 2022-07-26
Add config to mount a scratch directory

rendhalver:
This adds an optional config to mount a scratch directory into your ocm container.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -87,8 +87,8 @@ fi
 ### AWS token pull
 if [[ -d "${HOME}/.aws" ]]; then
   AWSFILEMOUNT="
--v ${HOME}/.aws/credentials:/root/.aws/credentials:ro 
--v ${HOME}/.aws/config:/root/.aws/config:ro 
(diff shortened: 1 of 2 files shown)


ocm-container.sh
```diff
@@ -120,6 +120,12 @@ if [ -d "${OPS_UTILS_DIR}" ]; then
 "
 fi
 
+### mount a scratch dir
+if [ -n "$SCRATCH_DIR" ]
+then
+  SCRATCH_DIR_MOUNT="-v ${SCRATCH_DIR}:/home/scratch/"
```
clcollins, line 126: Is this intended to be `/home/scratch` or `~/scratch` (/home/$USER/scratch)?

clcollins (commented):
One question about the mount point inside the container, but otherwise, LGTM.

Can you confirm where the scratch dir should be mounted?

ocm-container.sh
rendhalver: I mounted it there locally for a reason I can't remember.
I guess `~/scratch` makes more sense. We are mounting the utils dir under `~/` so let's move it there.
I will update the branch.

PR #127, opened 2022-08-05
Add flag to disable the ssh socket mount

rendhalver:
There may be better ways to handle this but I would like the option to disable it for now.


ocm-container.sh
```diff
@@ -157,7 +164,7 @@ ${AWSFILEMOUNT} \
 ${SSH_AGENT_MOUNT} \
 ${OPS_UTILS_DIR_MOUNT} \
 ${SCRATCH_DIR_MOUNT} \
--v ${HOME}/.ssh/sockets:/root/.ssh/sockets \
+-v /root/.ssh/sockets \
```
clcollins, line 167: This isn't a valid volume mount.  I'm assuming you meant to use the `$SOCKET_MOUNT`  var?

rendhalver: Ah whoops. I will fix that.

clcollins:
Excellent, thanks!
/lgtm
/approve

PR #129, opened 2022-08-09
Add omc to ocm-container

karthikperu7:
This PR adds the `omc` utility to ocm-container which is a very handy tool to inspect must-gathers. omc is better than omg as it is a lot more feature rich and close to using the `oc` cli against a must gather than the `omg` tool. You can read a short brief in the project's readme: https://github.com/gmeghnag/omc . @mrbarge also demo'd this in one of the APAC sprint demos.

I have tested the build and it works ok


karthikperu7:
we could consider removing `omg` utility in favour of this in a separate PR if required.

mrbarge:
:+1: from me, since switching to `omc` I've never looked back.

iamkirkbater:
/lgtm

Thanks for leaving out the removal of `omg`, we can potentially deprecate that but I'd like to wait so that we can communicate that to our users properly.

PR #130, opened 2022-08-11
Performance and Cluster Login bugfix

clcollins:
- Removes useless ocm list command
- Use OCM Backplane tunnel and logging for sre-login
- Update README.md to reflect current sre-login/automatic login


rendhalver:
This is rather great! Thank you!
/lgtm

PR #132, opened 2022-08-24
Checks to make sure backplane is installed for sre-login script

clcollins:
Backplane must be installed for sre-login to work, and there have been a number of questions in the last two or three weeks about this.  This PR adds a check for backplane and bombs out with a specific error message if backplane is not installed.

### DIFF
--- a/utils/bin/sre-login
+++ b/utils/bin/sre-login
@@ -1,6 +1,13 @@
 #!/bin/bash -e
 # OCM_CONTAINER_DOC: Logs into a cluster via the command line
 
+# Check that backplane is installed
+if ! ocm backplane version > /dev/null
+then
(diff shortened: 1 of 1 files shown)


utils/bin/sre-login
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash -e
 # OCM_CONTAINER_DOC: Logs into a cluster via the command line
 
+# Check that backplane is installed
+if ! ocm backplane version
```
iamkirkbater, line 5: We'd probably want to pipe the output of this to /dev/null, right?

rendhalver: Possibly. It seems to work okay without it though.
It may be handy seeing the error message in case something else is going wrong.

clcollins: Piped to dev null, thanks!

PR #137, opened 2022-10-10
Prevents unnecessary parent shell exit on failure; formatting fixes

clcollins:
Replacing the `exit 1` with `return` should prevent unnecessarily
closing the parent shell when skipping the rest of the sourcing due to
the missing `OFFLINE_ACCESS_TOKEN` variable.

The `exit 1` prevents the use of bash to troubleshoot issues in
partly-build ocm-contianer images or running the container image without
the bootstrapping built in for OCM.

Also, removes unnecessary `;` linebreak characters, and standardizes
if/elif/fi formatting.


iamkirkbater (commented):
/lgtm

PR #143, opened 2022-10-19
Clean up basic scripts

AlexVulaj:
Clean ups to bash files throughout the project after running them through ShellCheck.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -48,7 +48,7 @@ while [ "$1" != "" ]; do
          ;;
 
     * )
-      ARGS+=($1)
+      ARGS+=("$1")
       ;;
   esac
   shift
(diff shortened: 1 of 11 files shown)


AlexVulaj:
@clcollins I cherry-picked these changes over to @iamkirkbater 's [arm64 branch](https://github.com/iamkirkbater/ocm-container/tree/arm64) since I'm on an M1.  I was able to succesfully initialize, build, and connect to clusters in both stage and prod.

I definitely think we should have someone who isn't on an M1 test the changes directly from this PR as well before considering merging.

ocm-container.sh
```diff
@@ -48,7 +48,7 @@ while [ "$1" != "" ]; do
          ;;
 
     * )
-      ARGS+=($1)
+      ARGS+=("$1")
```
AlexVulaj, line 51: I didn't want to change it here in case I was missing something, but I found it odd that we take `ARGS` as an array here, especially when immediately below we exit out of the script with an error code if it has more than one element.  Is there a reason that `ARGS` needs to be an array?

utils/bin/pause-hive-sync
```diff
@@ -28,14 +28,12 @@ _logged_in(){
...
 
 # Cluster deployment must be provided
-while getopts ":hlc:s:" opt; do
+while getopts ":hc:s:" opt; do
   case ${opt} in
     h ) _help ; exit 0
       ;;
     c ) CLUSTER_DEPLOYMENT=${OPTARG}
       ;;
-    l ) LIST="true"
```
AlexVulaj, line 37: I couldn't find anywhere in the entire project that this was used, but please correct me if I'm missing something and I can easily add it back.

clcollins: I don't believe it's being used anymore either.  Must have been cruft from the past.

PR #146, opened 2022-11-16
include the osd-network-verifier utility

yithian:
since manually running the osd-network-verifier is now part of the investigation process for cluster provisioning delays, the network verifier should be part of ocm-container


fahlmant:
This is already included in `osdctl` under `osdctl network verify-egress`

yithian:
oh, i see. in that case https://github.com/openshift/osdctl/pull/286

PR #148, opened 2022-12-01
Fix `command not found` error during docker build

mrWinston:
Currently building ocm-container fails with the following error message:
```
[3/11] STEP 7/10: RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux" '$0~asset {print $2}' sha256sum.txt)"
/bin/bash: line 1: openshift-client-linux-arm64-4.11.16.tar.gz: command not found
Error: building at STEP "RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux" '$0~asset {print $2}' sha256sum.txt)"": while running runtime: exit status 127
```

The issue seems to be that the downloaded sha56sum.txt file now contains a line reading `openshift-client-linux-arm64-4.11.16.tar.gz` which breaks the mechanism by which the file name to download is being determined. This PR 'fixes' that by maching for `openshift-client-linux-4`so that `openshift-client-linux-arm64` is being excluded.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,9 +105,9 @@ WORKDIR /oc
 # Download the checksum
 RUN curl -sSLf ${OC_URL}/sha256sum.txt -o sha256sum.txt
 # Download the binary tarball
-RUN /bin/bash -c "curl -sSLf -O ${OC_URL}/$(awk -v asset="openshift-client-linux" '$0~asset {print $2}' sha256sum.txt)"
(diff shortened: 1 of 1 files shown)


rendhalver:
Hrm. Interesting.
This is also failing with the same error under `podman` on Fedora.

I only see this if I run the build with `--no-cache`.

mrWinston:
Yeah, the image cache doesn't get invalidated becuase nothing in the dockerfile or base image actually changed. So the error only comes up if you either disable the cache using the flag, or update the base image.

clcollins:
This is fixed by #149 

But thank you for your PR submission.  Apologies that I didn't notice it before working on this as well. :smile: 

/close

PR #149, opened 2022-12-06
Fixes oc binary selection

clcollins:
Linux arm64 binaries were recently added to the x86_64 client mirror
directory, breaking the selection of the proper binary for building
ocm-container.  This selects the binary properly with a more stringent
regex match.


iamkirkbater (commented):
/lgtm

PR #151, opened 2022-12-08
 Adds update to address CVEs and patches

clcollins:
A number of libraries (curl, gnupg, etc) have CVEs in the base image
that won't be patched until the next base image version is released, so
this adds an initial step to run a microdnf update as a first stage of
the build.

Also adds .github/* to .gitignore, to skip merging local github actions from personal branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 env.source
+.github/*
+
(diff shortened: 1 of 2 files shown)


clcollins:
/hold

Need to remove an accidentally committed github action

PR #152, opened 2022-12-09
ubi9

clcollins:
- Auto sync from upstream
- Update branch to master
- Use personal fork
- Upadate base image to ubi9-minimal


Dockerfile
```diff
@@ -1,96 +1,70 @@
...
+      procps-ng \
+      python3 \
+      python3-pip \
+      rsync \
+      sshuttle \
+      tar \
+      vim-enhanced \
+      wget \
+      && microdnf clean all \
+      && rm -rf /var/yum/cache
```
rendhalver, line 44: This may need that `\`.
The build succeeded using `--no-cache` without it so it may not be necessary.

clcollins: It won't need the \.  It's the last on the line.

rendhalver: That would explain why it built properly. :)

PR #157, opened 2023-01-17
[OSD-14655] Fix ocm-container.sh script

zmird-r:
ocm-container loads both files if .aws exists but it should verify if the files exists first.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -85,11 +85,13 @@ then
 fi
 
 ### AWS token pull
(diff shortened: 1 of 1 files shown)


T0MASD:
couple of minor suggestions for `then` syntax, but overall great logic to check for the actual aws cli conf files instead of assuming they're in `.aws` dir :+1: 
/lgtm

zmird-r:
I fixed the style. thank you @T0MASD

PR #162, opened 2023-02-11
Adds arm64/amd64 intercompatibility

iamkirkbater:
This PR adds full arm64/amd64 intercompatibility.

This has been designed so that there should be no need to change the way anything works.  It builds without extra arguments for the correct platform that you're running it on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,12 @@ FROM base-update as dnf-install
 # OCM backplane console port to map
 ENV OCM_BACKPLANE_CONSOLE_PORT 9999
 
+# Adds Platform Conversion Tool for arm64/x86_64 compatibility
+COPY utils/dockerfile_assets/platforms.sh /usr/local/bin/platform_convert
+
 # Add google repo
 COPY utils/dockerfile_assets/google-cloud-sdk.repo /etc/yum.repos.d/
+RUN platform_convert -i /etc/yum.repos.d/google-cloud-sdk.repo --x86_64 --aarch64
 
 RUN rpm --import https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-9 \
       && rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -104,12 +112,24 @@ ENV OCM_URL="https://api.github.com/repos/${OCM_URL_SLUG}/releases/${OCM_VERSION
 # ocm is not in a tarball
 RUN mkdir /ocm
 WORKDIR /ocm
+
+## arm-native
+# Download the checksum
+RUN [[ $(platform_convert "@@PLATFORM@@" --x86_64 --arm64) != "arm64" ]] && exit 0 || /bin/bash -c "curl -sSLf $(curl -sSLf ${OCM_URL} -o - | jq -r '.assets[] | select(.name|test("linux-arm64.sha256")) | .browser_download_url') -o sha256sum.txt"
```
iamkirkbater, line 118: I'm not a _huge_ fan of the way this is done, but these quotes inside single quotes inside quotes just absolutely _destroy_ any chance of shell expansion :(

Dockerfile
```diff
@@ -271,7 +318,8 @@ RUN jira completion bash > /etc/bash_completion.d/jira
 RUN oc completion bash > /etc/bash_completion.d/oc
 RUN ocm completion > /etc/bash_completion.d/ocm
 RUN osdctl completion bash > /etc/bash_completion.d/osdctl
-RUN rosa completion bash > /etc/bash_completion.d/rosa
+# rosa is only available for amd64 platforms so ignore it
+RUN [[ $(platform_convert "@@PLATFORM@@" --amd64 --arm64) != "amd64" ]] && rm ${BIN_DIR}/rosa || rosa completion bash > /etc/bash_completion.d/rosa
```
iamkirkbater, line 322: Reviewers take note - there will be NO `rosa` binary inside the arm64 version, that's the only binary that doesn't have a pre-built version.

geowa4: Personally, I think I would prefer a dummy script that prints out "`rosa` CLI not supported on arm64. Please do X instead." and then exits with a non-zero status code. The argument against that idea is any script that runs `command -v rosa` will succeed when its use is guaranteed to fail.

rendhalver:
I spun this up locally on my x86_64 Fedora machine and all the binaries are showing up as `ELF 64-bit LSB executable, x86-64`.
It's connecting to clusters as well and I can interact with apis so it all looks good to me.
LGTMing and holding for now. Feel free to take the hold off when you are ready to merge.

/lgtm
/hold

PR #164, opened 2023-02-16
[OSD-15271] skip osdctl version check when generating bash completion

T0MASD:
Since osdctl started to check for version, ocm-container build can get stuck due to required user input for `osdctl completion` command as seen from:

```
$ ./osdctl completion bash
The current version () is different than the latest released version (v0.14.0).
It is recommended that you update to the latest released version to ensure that no known bugs or issues are hit.
Please confirm that you would like to continue with [y|n]
```


clcollins:
/lgtm
/approve

I realize we should probably strive to make sure osdctl is up to date, but there are  definitely use cases for building with an older version.

PR #172, opened 2023-03-14
Adds optional PersonalizationFile config

iamkirkbater:
Over my last shift during the multitude of investigations I found myself wishing I had somewhere that automatically sourced some convenient aliases into ocm-container.  This adds that configuration.

The idea here being that each individual SRE can define their own set of helpful aliases rather than being forced to use the same list of them.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -5,11 +5,12 @@ usage() {
   usage: $0 [ OPTIONS ] [ Initial Cluster Name ]
   Options
   -e  --exec          		Path (in-container) to a script to run on-cluster and exit
+  -d --disable-console-port   	Disable automatic cluster console port mapping (Linux only; console port will not work with MacOS)
   -h  --help          		Show this message and exit
+  -n  --no-personalizations     Disables personalizations file, typically used for debugging potential issues or during automated script running
   -o  --launch-opts   		Sets extra non-default container launch options
   -t  --tag           		Sets the image tag to use
   -x  --debug         		Set bash debug flag
-  -d --disable-console-port   	Disable automatic cluster console port mapping (Linux only; console port will not work with MacOS)
 
   Initial Cluster Name can be either the cluster name, cluster id, or external cluster ID.
 EOF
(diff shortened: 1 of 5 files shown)


ocm-container.sh
```diff
@@ -179,6 +179,20 @@ then
   PORT_MAP_OPTS="--publish-all"
 fi
 
+if [ -n "$ALIAS_FILE" ]
+then
+  if [ -f "$ALIAS_FILE" ]
+  then
+    ALIASMOUNT="-v ${ALIAS_FILE}:/root/.config/alias.d/alias.sh"
+  elif [ -d "$ALIASFILE" ]
```
AlexVulaj, line 187: should this be `$ALIAS_FILE` here?

iamkirkbater:
Yeah, I was considering just using those, but sometimes I use those for development stuff, like if I want to add a specific directory as a volume mount for a specific reason (like, when I need to reset a bunch of AAO accounts I'll mount a directory just for that reset so I can store all the account names, etc).

The other thing I wonder with in that is that we'd still need to add _some_ code in order to have the bash shell automatically source the alias/shell files in order to pull them into the session.  So at a minimum we'd need to keep the 99-alias.bashrc file logic, and then anyone wishing to add that functionality would have to update their ocm config with:

```
OCM_CONTAINER_LAUNCH_OPTS="-v /path/to/alias.d:/root/.config/alias.d"
```

Which from a usability standpoint now means that the end-user needs to know exactly where the alias folder sits, assuming that it _won't_ be overwritten if you also use `ocm-container -o` during a run command.

Though, as I'm working through this comment, since this allows a directory, I'm wondering if instead of calling it an `ALIAS_FILE` we should call this something along the lines of a `PERSONALIZATION` option?  Since it's not limited to _just_ aliases.  Theoretically a user could put additional bash functions here, or other shell customizations like resetting their PS1 or something 🤔

ocm-container.sh
iamkirkbater:
```suggestion
  elif [ -d "$ALIAS_FILE" ]
```

T0MASD:
nice, +1 from me, pretty safe since it only sources files matching sh extention
/lgtm

PR #174, opened 2023-03-28
Adds the new public backplane cli

iamkirkbater:
This adds the new public backplane cli to ocm-container.

Builds from source right now because there's no consumable binary yet, I'm sure that will come in the near future.

Also currently depends on the feedback here: https://github.com/openshift/backplane-cli/pull/57/files#r1150515048, giving us the ability to mount an entire directory of configs.  Then, the difference is for ocm-container users we would have a `config.json` for prod, `config.staging.json` for staging (assuming you `OCM_URL=staging ocm-container`) and `config.integration.json` for integration, again assuming you `OCM_URL=integration ocm-container` for that environment.


iamkirkbater:
This doesn't have a _hard_ dependency on the backplane changes, as it works independently without it, we would just end up duplicating config locally or setting a local env var for working with backplane outside of the container.

clcollins:
This is still beta, I think?  It'd be worth holding off merging this into the main branch until there's been a little more testing, but maybe we can create a dev branch here for cases like these where we want to make it easy to test, but wouldn't want to make it the default for everyone until it's been tested more?

utils/bin/sre-login
```diff
@@ -64,6 +64,4 @@ cluster_listening=$(jq -r '.api.listening' <<< "$clusterjson")
 # Login to the Cluster
 
 echo "Cluster ID: $cluster_id"
-ocm backplane tunnel -D > tunnel.log
```
rendhalver, line 67: This is my favourite part of this MR.

rendhalver:
Getting a cluster console is not working with this and that's a blocker.
We need podman and a few other components from the old backplane for this to work properly.
We need to hold off merging this until console works.

iamkirkbater:
Closing this as the changes were incorporated as part of #187

PR #177, opened 2023-04-21
Add oc alias fn to append compliance reason

hectorakemp:
Adds an ergonomic shell expansion that allows this: 
`oc get po -A --reason 'compliance'`

to be expanded to this:
`ocm backplane elevate 'compliance' -- get po -A

Video: 
[Screencast from 2023-04-21 17-43-20.webm](https://user-images.githubusercontent.com/28124379/233574963-b30a426c-b274-4597-b710-0508c04579e7.webm)
To test:
Run `make build-ocm-container` in https://github.com/openshift/backplane-cli to use the latest version. Make sure that `.config/backplane/config.json` is set correctly (reach out to the backplane team for how to do this). 

Run the above command. The `Impersonate-Extra-Reason` header should get added to your oc request. 

Current drawbacks: 
Only works if `--reason` is appended to the end of a request. 

- [ ] Requires https://github.com/openshift/ocm-container/pull/174 to be merged.

### DIFF
--- a/utils/bashrc.d/06-sre-utils.bashrc
+++ b/utils/bashrc.d/06-sre-utils.bashrc
@@ -20,3 +20,33 @@ fi
 function cluster_function() {
   oc config view  --minify --output 'jsonpath={..server}' | cut -d. -f2-4
 }
+
+function oc() {
+    if [[ ! "$@" =~ "--reason" ]]
+    then
+        /usr/local/bin/oc $@
+        return $?
+    fi
+
+    let i=0
+    for arg in "$@"
+    do
+        if [[ $arg == "--reason" ]]
+        then
+            let reason_flag_pos=$i
+            let reason_text_pos=$i+1
+        fi
+
+        # get elevation text
+        if [[ $i == $reason_text_pos ]]
+        then
+            reason_text=$arg
+        fi
+
+        let i++
+    done
+
+    OC_POS_ARGS=( "${@:1:$reason_flag_pos}" )
+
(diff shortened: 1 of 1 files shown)


T0MASD:
Thanks this is really great to that splunk can handle the reason!
imho it would be best to update the `oc_elevate` flow (currently in gitlab) that every time you run `oc-elevate get po -A` inside the container, you get a prompt to enter a reason before command is executed
/hold

rendhalver:
Given this needs the new backplane it may be best to merge this after #174.
/hold

iamkirkbater:
I'm personally a `-1` on this, because it's not portable and for the users who don't know better they won't understand why `oc --reason` won't work outside of OCM Container.

I'm not sure what specific `oc_elevate` flow is in gitlab, but I could get behind some kind of plugin for `oc` that accepts `--reason` as a parameter.  In fact I think this would be pretty easy to do with the same code we have here - I think we'd just need to rename the file something like `oc_elevate` and put it in the PATH and it would automatically be found.

If an individual would still want to accomplish something like this then adding this to their personalizations file would be appropriate, I just dislike the idea of bulk-changing upstream functionality without being explicit as to who's changing that functionality.

hectorakemp:
Totally fine with me @iamkirkbater - could we explore something like an `oc-reason` script that gets put into `ocm-container`, so it doesn't conflict with anything that's already there? The main thing is just that it's awkward to use `ocm backplane elevate --reason ""` to attach the reason to the elevation. There's definitely space for something more ergonomic, I think.

iamkirkbater:
100% - I'm totally on board with that.

PR #187, opened 2023-07-04
Add Backplane dependencies for bp-cli 0.1.7

samanthajayasinghe:
Adding public backplane podman dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,11 +44,24 @@ RUN microdnf --assumeyes --nodocs install \
       tar \
       vim-enhanced \
       wget \
+      podman \
+      fuse-overlayfs \
+      crun\
       && microdnf clean all \
       && rm -rf /var/cache/yum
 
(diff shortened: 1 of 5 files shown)


T0MASD:
Please update the PR with the jira reference, if there's no jira please create one, this is very good change that deserves visibility

utils/containers.conf
```diff
@@ -0,0 +1,12 @@
+[containers]
```
rendhalver, line 1: It seems better to put this file into `utils/dockerfile_assets`

samanthajayasinghe: Move the file

samanthajayasinghe:
> Please update the PR with the jira reference, if there's no jira please create one, this is very good change that deserves visibility

Thanks, @T0MASD , Created a new JIRA ticket and add the reference

iamkirkbater:
/approve

I think there are still some changes that are requested, but I'll approve the direction of this PR and someone else can LGTM this when it's ready

PR #193, opened 2023-07-10
fix syntax error

yithian:
This fixes the below error

```
$ ocm-container hivep03uw1
/usr/local/bin/ocm-container: line 266: syntax error: unexpected end of file
```


rendhalver:
This was fixed in #192

PR #195, opened 2023-07-11
Add mount for managed notifications

rendhalver:
I keep needing to use these for service logs so adding a mount for them seems like a good plan


clcollins:
This is a public repo, as opposed to ops-sop ($OPS_UTILS_DIR).  Why not build it into the image and skip all the mounting business?

rendhalver:
> This is a public repo, as opposed to ops-sop ($OPS_UTILS_DIR). Why not build it into the image and skip all the mounting business?
> 
> Additionally, this is only useful with `osdctl` which also parses the template from URL, so does it need to be local at all?

Very good points. Thanks for the sanity check Chris.

I think the only reason this would be used would be for sending edited copies of the ones in managed-notifications and that's probably just encouraging bad behavior.
I am gonna stop myself and others from doing that and close this.

PR #197, opened 2023-07-12
Adds check for SSH_AUTH_SOCK

clcollins:
Checks that the SSH_AUTH_SOCK env variable is set, and prompts to run
ssh-agent if not.  This should have no effect on anyone running
ocm-container locally for the most part, and provide a friendly prompt
for folks running ocm-container via ssh to another machine.

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -84,6 +84,11 @@ source ${OCM_CONTAINER_CONFIGFILE}
 ### SSH Agent Mounting
 operating_system=`uname`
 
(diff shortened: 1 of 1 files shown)


iamkirkbater:
This has no effect on an M1 Mac running podman machine, everything seems to line up properly here. 

Though without the need to connect to bastions anymore I do wonder if this SSH agent passthrough functionality is even required anymore?

I won't LGTM this yet to give some time for other discussion or use cases to be tested, but I will LGTM this later this week if there's no other discussion needed for this - I have no problems with adding this and coming back to the discussions about it later.

PR #199, opened 2023-07-14
add oc-elevate function

T0MASD:
prompts for justification before commands as cluster admin

example when providing a justification:
```
[~ {production} (hive-xxx:default)]$ oc-elevate get pod -A -o json | jq -r '.items | length'
Enter justification for performed elevated action (60s timeout):test test
445
```

example when not providing a justification due to timeout or empty string
```
[~ {production} (hive-xxx::default)]$ oc-elevate get pod -A -o json | wc -l
Enter  justification for performed elevated action (60s timeout):
No justification for elevated action was given exiting...
0
```

Confirmed with security team, justification is sent to the audit log

### DIFF
--- a/utils/bashrc.d/12-oc.bashrc
+++ b/utils/bashrc.d/12-oc.bashrc
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+oc-elevate() {
+  # set timeout and messaging
+  TIMEOUT="60"
+  PROMPT_TEXT="Enter justification for performed elevated action (${TIMEOUT}s timeout):"
+  NOOP_TEXT="No justification for elevated action was given exiting..."
+
+  # wait for user input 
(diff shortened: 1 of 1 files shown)


clcollins:
This is awesome, and LGTM generally.  Would you be opposed to the function taking an argument to add the REASON via a flag using bash's GETOPS (or something), to support just running it with a flag, or prompting if run without one?

Something like `oc-elevate -r "This was my reason for elevating"`?

clcollins:
Oh, and one more idea - this is already named properly and would be an excellent candidate for an OC plugin - it just needs to go into `/usr/local/bin` instead of the bashrc.d utils, and execute the function  automatically.

`oc elevate get pod -A....`

iamkirkbater:
+1 to having a reason passed in and not _have_ to provide it in a text prompt afterwards.

PR #205, opened 2023-07-21
Mount local certificate authority anchors

T0MASD:
Mounts certificate authority trust source to avoid self-signed certificate errors

before
```
$ curl https://someplace.redhat.com
curl: (60) SSL certificate problem: self-signed certificate in certificate chain
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

after
```
curl https://someplace.redhat.com
<html><body>You are being <a href="https://someplace.redhat.com/users/auth/geo/sign_in">redirected</a>.</body></html>```
```

### DIFF
--- a/env.source.sample
+++ b/env.source.sample
@@ -64,3 +64,7 @@ here \
 ### In the case of a directory being defined, all *.sh files in that directory
 ### will be `source`'d.
 #export PERSONALIZATION_FILE=/home/myuser/path/to/my/defined-personalizations.sh
+
+### CA_SOURCE_ANCHORS
+### The CA_SOURCE_ANCHORS is a local trusted certificate authorities to mount and use in a container.
+#export CA_SOURCE_ANCHORS=/etc/pki/ca-trust/source/anchors
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -81,6 +81,13 @@ fi
 
 source ${OCM_CONTAINER_CONFIGFILE}
 
+### Mount certificate authority trust source to avoid self-signed certificate errors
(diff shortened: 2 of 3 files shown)


clcollins:
Can we just add the CAs to the image?  They're not secret, and it would avoid yet another mount of our local filesystem into the container.  This would allow any users to use OCM container without having to have the CAs, too, which I think is an excellent advantage of the project.  I like being able to just pull the image and go, without needing to customize my laptop at all to get things to work :)

clcollins:
Coming back to this, I just added this to another image I use - would you be open to making this change in the Dockerfile instead of adding the mount, @T0MASD:

```
RUN curl -sSL -o /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem
RUN curl -sSL -o /etc/pki/ca-trust/source/anchors/2022-IT-Root-CA.pem https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem
RUN update-ca-trust
```

T0MASD:
@clcollins unfortunatley those certs aren't public :( 
```
$ curl https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem
curl: (6) Could not resolve host: certs.corp.redhat.com
```

clcollins:
Ah, bummer.  Maybe we can grab those in some other way in the future.  

/lgtm
/approve

PR #207, opened 2023-07-27
Add settings for repo and tag

rendhalver:
Adding an way to configure which repo to use and an extra way to set tag using env vars

I have been using direnv as a way to set tag and repo for ocm-container this method will also work with env.source so I thought I would share it.
We currently don't have a way to configure which repo to use so this will aslo allow us to use the pre-built images on quay.

### DIFF
--- a/env.source.sample
+++ b/env.source.sample
@@ -1,6 +1,12 @@
 ### Autodetect container subsystem
 export CONTAINER_SUBSYS=${CONTAINER_SUBSYS:-"$(which podman >/dev/null 2>&1 && echo podman || echo docker)"}
 
+### Set the container repo and tag we want to use.
+### Uncomment this to use the pre-built images Set it to localhost to use local images
+#export OCM_CONTAINER_REPO="quay.io/app-sre"
+### set this to the tag you wish to use
+export OCM_CONTAINER_TAG="latest"
+
 ### Select cli, options include:
 ### - "ocm"
 ### - "rosa"
--- a/env.source.sample
+++ b/env.source.sample
@@ -2,8 +2,9 @@
 export CONTAINER_SUBSYS=${CONTAINER_SUBSYS:-"$(which podman >/dev/null 2>&1 && echo podman || echo docker)"}
 
 ### Set the container repo and tag we want to use.
-### Uncomment this to use the pre-built images Set it to localhost to use local images
-#export OCM_CONTAINER_REPO="quay.io/app-sre"
+### Uncomment this to use local images
+### The default is repository `quay.io/app-sre`
+#export OCM_CONTAINER_REPO="localhost"
 ### set this to the tag you wish to use
 export OCM_CONTAINER_TAG="latest"
 
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -9,6 +9,7 @@ usage() {
   -h  --help          		Show this message and exit
   -n  --no-personalizations     Disables personalizations file, typically used for debugging potential issues or during automated script running
   -o  --launch-opts   		Sets extra non-default container launch options
+  -r  --repo                    Sets the container repository to use (We use quay.io/app-sre by default)
   -t  --tag           		Sets the image tag to use
   -x  --debug         		Set bash debug flag
 
@@ -46,6 +47,9 @@ while [ "$1" != "" ]; do
                                     ;;
     -t | --tag )                    shift
                                     BUILD_TAG="$1"
+                                    ;;
+    -r | --repo )                   shift
+                                    BUILD_REPO="$1"
                                     ;;
     -x | --debug )                  set -x
                                     ;;
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -17,16 +17,8 @@ EOF
 }
 
 ARGS=()
-if [[ -n "${OCM_CONTAINER_TAG}" ]]; then
-  BUILD_TAG="${OCM_CONTAINER_TAG}"
-else
-  BUILD_TAG="latest"
-fi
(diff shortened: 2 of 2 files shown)


clcollins:
/hold 

For discussion in #forum-ocm-container about if this is needed or not

iamkirkbater:
I like the idea. I'm +1 to this - but also only if we add a `-i | --image` flag to the CLI so that you can change it on each invocation _without_ using env vars.  We already have the similar functionality for `-t | --tag` that I use heavily when developing and/or A/B testing images locally, so it makes sense to expand this functionality to the image itself.

ocm-container.sh
```diff
@@ -17,7 +17,16 @@ EOF
...
+if [[ -n "${OCM_CONTAINER_TAG}" ]]; then
+  BUILD_TAG="${OCM_CONTAINER_TAG}"
+else
+  BUILD_TAG="latest"
+fi
+if [[ -n "${OCM_CONTAINER_REPO}" ]]; then
+  BUILD_REPO="${OCM_CONTAINER_REPO}"
+else
+  BUILD_REPO="localhost"
+fi
```
iamkirkbater, line 29: I very-much dislike `else` blocks unless they're absolutely needed.  Can we potentially either use bash parameter substitution logic here like 

```
BUILD_TAG="${OCM_CONTAINER_TAG:=latest}"
```

 or if we're worried about that functionality we could do something like

```
BUILD_TAG="latest"
BUILD_REPO="localhost"

if [[ -n "${OCM_CONTAINER_TAG}" ]]; then
  BUILD_TAG="${OCM_CONTAINER_TAG}"
fi
... etc
```

env.source.sample
```diff
@@ -1,6 +1,12 @@
 ### Autodetect container subsystem
 export CONTAINER_SUBSYS=${CONTAINER_SUBSYS:-"$(which podman >/dev/null 2>&1 && echo podman || echo docker)"}
 
+### Set the container repo and tag we want to use.
+### Uncomment this to use the pre-built images Set it to localhost to use local images
```
iamkirkbater, line 5: Can we say here what the default is if this is not set since we're providing a value that's not the default? - looks like we're using `localhost` as the default in the code - though I also have a comment about that as well in that file.

ocm-container.sh
```diff
@@ -17,7 +17,16 @@ EOF
...
-BUILD_TAG="latest"
+if [[ -n "${OCM_CONTAINER_TAG}" ]]; then
+  BUILD_TAG="${OCM_CONTAINER_TAG}"
+else
+  BUILD_TAG="latest"
+fi
+if [[ -n "${OCM_CONTAINER_REPO}" ]]; then
+  BUILD_REPO="${OCM_CONTAINER_REPO}"
+else
+  BUILD_REPO="localhost"
```
iamkirkbater, line 28: Should we default this to `quay.io/app-sre` instead of localhost?

iamkirkbater (changes requested):

ocm-container.sh
rendhalver: Aha! Thanks Kirk!
I was trying to work out how to do it that way!
I will get that fixed. :)

ocm-container.sh
rendhalver: I am also thinking that's a good idea.

rendhalver:
> I like the idea. I'm +1 to this - but also only if we add a `-i | --image` flag to the CLI so that you can change it on each invocation _without_ using env vars. We already have the similar functionality for `-t | --tag` that I use heavily when developing and/or A/B testing images locally, so it makes sense to expand this functionality to the image itself.

Should we use image or just use `-r | --repo` to set the container repo and leave the image to ocm-container?

env.source.sample
rendhalver: Good call.

iamkirkbater:
> Should we use image or just use -r | --repo to set the container repo and leave the image as ocm-container?

I could go either way on this - `--repo` would work and is already implemented, but `--image` could also be just as flexible, as someone could pass whatever repo/image name they wanted there while it would default to `localhost/ocm-container`.

I won't block the PR on this point, but I think `--image` would be more flexible in the future without locking us into a "contract" with the `--repo` flag in the future if we ever want to change the name of the image itself, as removing that flag would be a breaking change and could cause UX disruption.

clcollins:
I thought this PR was not needed due to the discussion here:  https://redhat-internal.slack.com/archives/C01E0BC9550/p1690501322328629

This is adding extra steps and code we need to maintain when the user can just pull the image and be done with it.

rendhalver:
> I thought this PR was not needed due to the discussion here: https://redhat-internal.slack.com/archives/C01E0BC9550/p1690501322328629
> 
> This is adding extra steps and code we need to maintain when the user can just pull the image and be done with it.

I agree that there other ways to achieve this. I have tried a few and this seemed like the best way to do it without impacting how other people use ocm-container.

Maybe some explanation on how I am using this will help.
I am a firm believer in speeding up how I do things.

I use [direnv](https://direnv.net/) a lot to automate a bunch of stuff I do.
ocm-container is one of the things I use it for.
I am using direnv to set the tag to use based on which directory I am in or globally if I want to run a test version for a while to see how it runs. I tried doing this with aliases but it's clunky and requires reloading my shell or reloading my bash aliases each time I change it, direnv let's me do this without doing reloading.

As for future maintenance, if there is a more efficient way to do this I am open to suggestions.
The main bit I require is a way to set tags and repos with env vars. I am not concerned with how that works.

rendhalver:
/close
This will not be needed once our golang cli is done.

PR #216, opened 2023-09-01
Update osdctl version

iamkirkbater:
Updates osdctl version to be the latest


rendhalver:
From what I can tell the container build for this hasn't worked.
I can not see a tag for it.

PR #218, opened 2023-09-05
add dig utility

yithian:
Since we're doing so much network diagnosis, having more network tools built into ocm-container is useful. This commit adds dig via the bind-utils package


T0MASD:
+1
/lgtm

PR #232, opened 2023-10-31
Get backplane latest version

samanthajayasinghe:
## Why 
We have to update docker file to get the backplane's latest version. 
This PR will point to the latest backplane version always.


T0MASD:
we currently don't use `latest` in ocm-container build,  backplane tools together with periodic build should solve this

PR #235, opened 2023-11-06
don't get the hypershift binary if on arm64

iamkirkbater:
Allows the container to build for non-x86_64 machines.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -383,6 +383,7 @@ RUN chmod -R +x /out
 FROM quay.io/acm-d/rhtap-hypershift-operator as hypershift
 RUN mkdir -p /out
 RUN cp /usr/bin/hypershift /out/hypershift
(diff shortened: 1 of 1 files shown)


T0MASD:
tested on arm64 machine, however I get a build error:
```
--> 428c0e5aa71f
[16/16] STEP 25/40: RUN ocm backplane completion bash > /etc/bash_completion.d/ocm-backplane
--> f220c88293a5
[16/16] STEP 26/40: RUN hypershift --version
qemu-x86_64-static: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
Error: building at STEP "RUN hypershift --version": while running runtime: exit status 255

```

it appears binary is not overwritten
```
[~ {production} ]$ du -h /usr/local/bin/hypershift
127M	/usr/local/bin/hypershift
```

I ran the build on a system with no images

rendhalver:
I believe this is contained within #245.
/hold

PR #236, opened 2023-11-13
Add a symlink to /usr/local/bin for all files installed via backplane tools.

bergmannf:
/root/.local/bin/backplane/* is not on the path when using podman exec.

As I often run command via emacs TRAMP (or just in pipelines), this is will make those workflows work again.


T0MASD:
/close 
this is fixed in !237

PR #237, opened 2023-11-14
backplane tools builder

T0MASD:
adds building backplane tools as a separate step and explicitly copies the binaries. One weird thing with aws-cli :man_facepalming: is that it's weird python code with linking to c library so I follow aws install instructions placing that whole mess to /usr/local/aws-cli, in addition we also use a wrapper that sets proxy for aws cli hence the script in bashrc.d 
this fixes issue brought up in  https://github.com/openshift/ocm-container/pull/236


typeid (approved):
/lgtm

PR #239, opened 2023-11-19
add advanced upgrade service (AUS) CLI

geoberle:
the AUS CLI utility is an OCM CLI plugin to manage AUS upgrade policies on clusters and organizations in OCM

https://source.redhat.com/groups/public/sre/wiki/advanced_upgrade_service_aus

adding this to `ocm-container` is beneficial for AUS users, since they get the AUS plugin bundled with an up to date OCM utility. this way `ocm-container` can be used in CI systems where they validate their AUS policies and apply them after merge.

part of https://issues.redhat.com/browse/APPSRE-8538

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -228,6 +228,37 @@ RUN mkdir -p /out
 RUN cp /usr/bin/hypershift /out/hypershift
 RUN chmod -R +x /out
 
+
+FROM builder as aus-builder
+# Add `aus` utility for interacting the advanced upgrade service
+# https://source.redhat.com/groups/public/sre/wiki/advanced_upgrade_service_aus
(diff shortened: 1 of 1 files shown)


clcollins:
Hi @geoberle, thanks for the PR!

Who is the intended audience for the AUS utility?  They majority of users of ocm-container are likely SREP, who don't really have anything to do with scheduling upgrades.  However, I think it's fair to add utilities others might find helpful as well.  We just want to balance how much may need to be in the "core" ocm-container, and what might be added by other teams using ocm-container as a base image to customize their own.

What are your thoughts on this?

clcollins (commented):
Docker build works fine and AUS command runs as expected inside container built from the image.  Generally /lgtm, just awaiting more discussion from my comment above.

PR #241, opened 2023-11-23
handle better ocm-container config file creation

T0MASD:
when config not found, download one from upstream, promt for ocm token for 5min, if provided, update it and continue as normal

```
[tomas@ugnis ocm-container]$ rm ~/.config/ocm-container/env.source 
[tomas@ugnis ocm-container]$ ./ocm-container.sh 
Cannot find config file at /home/tomas/.config/ocm-container/env.source
Downloading sample config from upstream...
Paste your ocm token from https://cloud.redhat.com/openshift/token: dummy token ftw
Thanks, updating the sample config...
```
so the token is saved
```
[tomas@ugnis ocm-container]$ grep TOKEN ~/.config/ocm-container/env.source 
export OFFLINE_ACCESS_TOKEN="dummy token ftw"
```

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -74,9 +74,19 @@ export OCM_CONTAINER_CONFIGFILE="$CONFIG_DIR/env.source"
 
 if [ ! -f ${OCM_CONTAINER_CONFIGFILE} ]; then
     echo "Cannot find config file at $OCM_CONTAINER_CONFIGFILE";
-    echo "Run the init.sh file to create one."
-    echo "exiting"
-    exit 1;
+    echo "Downloading sample config from upstream..."
+    mkdir -p ${CONFIG_DIR}
+    curl -s https://raw.githubusercontent.com/openshift/ocm-container/master/env.source.sample --output ${OCM_CONTAINER_CONFIGFILE}
+    read -t 300 -p 'Paste your ocm token from https://cloud.redhat.com/openshift/token: ' CONFIG_OCM_TOKEN
+    if [[ $? -gt 128 ]]
+    then
+      echo -e "\nTimeout waiting for ocm token"
+      rm ${OCM_CONTAINER_CONFIGFILE}
+      exit 1
+    else
+      echo "Thanks, updating the sample config..."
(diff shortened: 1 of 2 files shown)


ocm-container.sh
```diff
@@ -74,9 +74,19 @@ export OCM_CONTAINER_CONFIGFILE="$CONFIG_DIR/env.source"
 
 if [ ! -f ${OCM_CONTAINER_CONFIGFILE} ]; then
     echo "Cannot find config file at $OCM_CONTAINER_CONFIGFILE";
-    echo "Run the init.sh file to create one."
-    echo "exiting"
-    exit 1;
+    echo "Downloading sample config from upstream..."
+    mkdir -p ${CONFIG_DIR}
+    curl -s https://raw.githubusercontent.com/openshift/ocm-container/master/env.source.sample --output ${OCM_CONTAINER_CONFIGFILE}
+    read -t 300 -p 'Paste your ocm token from https://cloud.redhat.com/openshift/token: ' CONFIG_OCM_TOKEN
```
rendhalver, line 80: I am thinking it would be good to do this if the configfile exists but the token is still set to `YOUR_TOKEN_HERE`.

T0MASD: it wouldn't happen as a result of this workflow, because if the token is not specified in the 5min window, the downloaded file is deleted and ocm container exits 
if the user downloads the config manually and places it there, setting the config is on them

on a side note, if we want to check the validity of the ocm token, we should probably source the config and verify the character length of the `OFFLINE_ACCESS_TOKEN` variable

ocm-container.sh
```diff
@@ -74,9 +74,19 @@ export OCM_CONTAINER_CONFIGFILE="$CONFIG_DIR/env.source"
 
 if [ ! -f ${OCM_CONTAINER_CONFIGFILE} ]; then
     echo "Cannot find config file at $OCM_CONTAINER_CONFIGFILE";
-    echo "Run the init.sh file to create one."
```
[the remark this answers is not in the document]
T0MASD: it's a second step in this direction (first being quay.io/app-sre/ocm-container ) where in the end ocm-container script is all that user needs to run it. The next part will be prompting for backplane endpoint, proxy address and arn for backplane config when script detects backplane config for the env is missing

PR #245, opened 2023-12-08
Containerfile updates

iamkirkbater:
* ROSA cli now has an arm64 build, so let's run the completion bash
* Hypershift CLI tool does NOT have an arm64 build, so we remove it from the container so this builds on arm64
* Install the session manager plugin for the aws CLI
* Updates the platform_convert script to work with a custom `64bit` tag for the aws linux ssm plugin rpm url.


rendhalver:
Thanks Kirk!
/lgtm

PR #250, opened 2024-01-02
add yum-utils and disable google repo

T0MASD:
add yum-utils and disable google repo after gcloud install as it is often source of timeouts when building on top of ocm-container in the future

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@
 ARG BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:9.1.0
 FROM ${BASE_IMAGE} as base-update
 
-RUN microdnf --assumeyes --nodocs update \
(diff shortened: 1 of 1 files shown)


iamkirkbater:
Are these related to daily ops tasks? I think adding these tools might be a bit too far of a deviation from the spirit of ocm-container, which is mainly to provide a stable, reproducible environment for SREs to access clusters and perform operations-related tasks.

I think running terraform tasks and packer operations falls outside of that scope.

I'm putting a hold on this for further discussion though.

/hold

T0MASD:
I've updated the PR to essentially disable google-cloud-sdk repo after the container build as it is often source of time outs for follow up builds

clcollins:
/lgtm

With the yum-utils and google-sdk changes, minus the others. :+1:

PR #256, opened 2024-03-01
adds buildx command to docker multi-platform builds

iamkirkbater:
Still need to address the 403 errors from Github, but that's a whole pile of sticks to address so let's start with just getting Docker to actually pull the correct images and see if that might start working on the nightly builds/pr checks.

Removed the caching in PR checks. We might need to refactor this because if we attempt to run multiple PR checks in quick succession this will hit the GH rate limiting. However, if we hit the rate-limiting and the build fails, it will cache that 403 response in the image layer and fail the build later.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -78,19 +78,6 @@ PLATFORM_BUILD_ARG="build"
 ### Start multi-platform build handling
 if [[ -n $PLATFORM ]] 
 then
-  #arch="$(echo $PLATFORM | cut -d/ -f2)"
-  #system_arch=""
-  #case "$(uname -m)" in
-  #  aarch64) system_arch="arm64";;
-  #  armv8b) system_arch="arm64";;
-  #  armv8l) system_arch="arm64";;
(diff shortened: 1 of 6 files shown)


build.sh
```diff
@@ -68,12 +74,37 @@ source ${OCM_CONTAINER_CONFIG}
 ### start build
 echo "Using ${CONTAINER_SUBSYS} to build the container"
 
+PLATFORM_BUILD_ARG="build"
+### Start multi-platform build handling
+if [[ -n $PLATFORM ]] 
+then
+  #arch="$(echo $PLATFORM | cut -d/ -f2)"
```
rendhalver, line 81: Do we need all of this here?

iamkirkbater: Removed. Thanks!

clcollins:
/hold

Dealing with other platform/arch issues ATM

iamkirkbater:
I'm going to close this as this will be fixed as we move to Konflux.

PR #262, opened 2024-04-03
adds ability to name container when cluster id provided

iamkirkbater:
Renames the container after startup to the display name of the cluster.

// still need to test with really long container names

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -192,14 +192,40 @@ then
   SCRATCH_DIR_MOUNT="-v ${SCRATCH_DIR}:/root/scratch/"
 fi
 
+## Set ocm url
+if [ -z $OCM_URL ]
+then
+  OCM_URL="production"
+fi
+
 ### Automatic Login Detection
 if [ -n "$ARGS" ]
 then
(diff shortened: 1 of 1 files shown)


clcollins:
I like this as a feature, but would prefer to keep as much of `ocm` inside the container.  I know there's already a usage of `ocm` outside for the persistent cluster histories, but I think it would be best to limit that and remove the ocm from outside if possible.

I don't have any other issues with it.  The renaming is cool, and the idea overall is great :)  

This might be more doable with the go-rewrite, if we could possibly wait a bit for that.  I'm hoping to have that available to test by the end of the week.

iamkirkbater:
That's a fair concern - the way this is written it should "fail closed" - though I may have to do some additional testing about not having the ocm binary installed and validate the login command worked before continuing.

Ideally - if this is NOT available the correct behavior would be to fall back to default functionality and NOT care. I'll double check this edge case.

clcollins:
> if this is NOT available the correct behavior would be to fall back to default functionality and NOT care.

Totally, but I think once we start calling stuff from the outside, folks are going to develop for it, and stop checking.  We're gonna end up with two environments.  "You get feature set X if you install ocm outside, but only a subset of the features if you use ocm-container as a container".  It's just going to pull more of the stuff outside of the container.

PR #263, opened 2024-04-03
Adds an option to run in the background in a 'headless' mode

iamkirkbater:
Adds a background option that allows the container to start the container and log into the cluster but not immediately attach to the session.

These podman sessions will need to be manually closed when you are done with them.


clcollins:
Fixed by #265

PR #264, opened 2024-04-05
BACKPLANE_CONFIG_MOUNT as rw because cluster-console attempt to write a config there

T0MASD:
Since we now use single backplane config for all envs, it should be ok to mount the bp conf location as rw, especially since cluster-console command attempts to write a temp nginx config there...

### DIFF
--- a/ocm-container.sh
+++ b/ocm-container.sh
@@ -246,7 +246,7 @@ then
 fi
 
 ## Set the mount path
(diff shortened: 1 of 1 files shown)


rendhalver:
/hold
Chris is working on moving the cli into Golang and he has pushed the console parts into Golang as well.
Let's hold off on this until we get that merged.

clcollins:
Fixed by #265

PR #265, opened 2024-04-10
Rewrites ocm-container in golang

clcollins:
This is a large PR, rewriting (or reimplementing, really) ocm-container in golang.  The intent is to make this more testable, more extensible, to fix some bash-induced bugs, and finally, to allow us to make and publish a binary to more easily, reliably and securely onboard new users.

This rewrite is an (almost) 1-1 feature parity and backward-compatible replacement for the ocm-container.sh file, as well as a number of other related tweaks. 

## Opinionated changes

In a (very) few places, it's a little more opinionated and will possibly require a minor change in usage, but for the most part opinionated changes are displayed as deprecation messages, and we can decide in the future how and when to handle them.  In most of these cases, it's to fix some ocm-container-isms that prevent us from utilizing some of the features of the container:

eg: `ocm-container CLUSTERID` is difficult to parse, and prevents us from easily using container-standard Entrypoint and CMD flags.  This PR deprecates that and adds `ocm-container --cluster-id CLUSTERID [--entrypoint=ENTRYPOINT] -- [CMD]` so that we can run arbitrary commands in ocm-container.  The original usage still currently works, but is marked Deprecated, and `--cluster-id` is compatible with other standard flags across other tools in use by our team.

In other cases, deprecation messages are included for things that will actually be deprecated, such as the use of OCM's OFFLINE_ACCESS_TOKEN.

## Feature Parity

I've tried as best as I can to have full 1:1 feature parity with the old shell script.  EG: PagerDuty tokens, Personalizations, etc should work exactly the same, and be configured exactly the same.  In some few cases, there may be a deprecation warning, to allow us to transition to a cleaner way of enabling these using the enhanced capabilities in golang and [https://github.com/spf13/viper](Viper/Cobra) configuration.

All features are enabled by default, but some may require additional configuration (the same configuration required for the feature in ocm-container.sh right now).  Any given feature may be disabled by using the `--no-FEATURENAME=true` CLI flag for testing, or setting the `no_featurename: true` config or related ENV var.

Note that env vars are prefixed with `OCMC_` to prevent accidental parsing of system or other env vars (eg: USER).

## Testing and Adoption

First, unset your ocm-container alias, if you have one.  The binary may be executed directly.

Get the binary: For testing this PR (not adoption or production), the `make build-snapshot` command will use Goreleaser to build a single OS/Arch appropriate binary for you to test with.

The image: ocm-container will by default use the upstream nightly build from quay.io/app-sre/ocm-container:latest.  No image build is required.  You may specifiy your own image.  See `ocm-container --help` for details.

Initial setup:

See the README.md in this PR for a thorough documentation on how configuration and migration works.   

The most basic setup can be configured via:

```
# CONTAINER_ENGINE may be podman or docker
ocm-container configure set engine CONTAINER_ENGINE
ocm-container configure set offline_access_token OCM_OFFLINE_ACCESS_TOKEN
```

Optionally, existing ocm-container users can make use of a [migration command to convert the env.source to a Viper ocm-container.yaml file](https://github.com/clcollins/ocm-container/blob/golang/README.md#migrating-configuration-from-the-bash-based-ocm-containersh-and-envsource-files).

Long term, I plan to expand this `ocm-container configure init` command to prompt the user to add the values to setup ocm-container and the desired features.

*** 

I would like to get as many folks involved in reviewing this and testing the functionality as possible, so we can cover all our bases and make this a seamless transition.

I also propose making a v0.1 tag for the existing master branch, and creating a pre-release v0.2 tag and binary for the contents of this PR when it merges into master, to mark the change in compatibility going forward.  However, there's no concern about merging this into master, as existing folks have their clone, and the repository DOES NOT need to be cloned to get and use the Go binary, so it can be tested in parallel with the old workflow.

Comments and reviews welcome.  Testing of the binary is especially!  Thanks in advance!


iamkirkbater:
This is great Chris! I'm super excited to test this.

Just a few notes from my read-through of your description: I absolutely appreciate the messaging to deprecate some functionality - however I'd like to consider this a breaking major version change with semver (so instead of tagging as v0.1 -> v0.2, we make this v0.1 -> v1.0) and I'm more OK with removing the functionality of most things without deprecation warnings when we go to publish this. 

As I would consider this PR a release candidate - I'm more than happy to keep the deprecation warnings though to help with the testing transition, but upon release I'd much rather have these deprecation warnings and the functionality be actually removed and call it a breaking change, rather than having to plan to actually remove later.

Aside from that - I'm excited to get this pulled down and start testing this during my shift next week :D

rendhalver:
> Just a few notes from my read-through of your description: I absolutely appreciate the messaging to deprecate some functionality - however I'd like to consider this a breaking major version change with semver (so instead of tagging as v0.1 -> v0.2, we make this v0.1 -> v1.0) and I'm more OK with removing the functionality of most things without deprecation warnings when we go to publish this.
> 
> As I would consider this PR a release candidate - I'm more than happy to keep the deprecation warnings though to help with the testing transition, but upon release I'd much rather have these deprecation warnings and the functionality be actually removed and call it a breaking change, rather than having to plan to actually remove later.

I like Kirk's logic here.
I am all for this option.

pkg/engine/engine.go
```diff
@@ -0,0 +1,347 @@
...
+
+func New(engine string, dryRun, verbose bool) (*Engine, error) {
+	e := &Engine{
+		dryRun:  dryRun,
+		verbose: verbose,
+	}
+
+	if verbose {
+		fmt.Println("using container engine:", engine)
+	}
```
iamkirkbater, line 81: I'd like to put stuff like this behind a logging library, if possible.

pkg/engine/engine.go
```diff
@@ -0,0 +1,347 @@
...
+	var err error
+
+	command := e.engine
+	args = append([]string{subcommand}, args...)
+
+	c := exec.Command(command, args...)
+
+	if e.verbose && !e.dryRun {
+		fmt.Printf("executing command: %+v\n", c)
+	}
```
iamkirkbater, line 111: Logging Library again.

pkg/ocmcontainer/envs.go
```diff
@@ -0,0 +1 @@
+package ocmcontainer
```
iamkirkbater, line 1: I might have missed it, but what is this file doing?

iamkirkbater (commented):
I've opened a slack thread to see if we can get a new branch in this repo so that we can merge this as-is and start making incremental improvements to this as we test it.

This would allow us to do things like address these rogue `fmt.Println` statments and put them behind a proper logger to hide them normally, etc, most importantly to not make this land on YOUR shoulders alone.

There's a bunch of great work here and I'd like to start iterating on it immediately, but I also don't want all of that work to fall on you, or bottleneck by having this roll through your personal fork first to merge into this one PR.

--

I do have one specific question about a file that just has a single package line in it though if you could clarify that. Thanks!

Again, great work here! I've pulled this down and plan on using it during my shift today to start seeing which pieces might need to be addressed sooner rather than later.

pkg/ocmcontainer/envs.go
clcollins: Nothing when you commented.  It's handling environment variables now. :D

README.md
```diff
@@ -1,165 +1,258 @@
...
+
+* No additional configuration required, other than on first-run (see below):
+* Can be disabled with `no_jira: true`
 
 Generate a Personal Access Token by logging into JIRA and clicking your user icon in the top right of the screen, and selecting "Profile". Then Navigate to "Personal Access Tokens" in the left sidebar, and generate a token.
 
-If you wish to use the `jira` cli outside of ocm-container, add the following to your environment by adding it to your .zshrc or .bashrc file. If you only wish to use this within ocm-container, you can also add the following to your `.env.source` file:
+If this is your first time using the JIRA CLI, ensure that the config file exists first with `mkdir -p ~/.config/pagerduty-cli && touch ~/.config/pagerduty-cli/config.json`. You'll also need to mount the JIRA config file as writeable by setting the `jira_dir_rw: true` configuration (or `export OCMC_JIRA_DIR_RW: true`) the first time. Once you've logged in to ocm-container, run `jira init` to do the initial setup.
```
rendhalver, line 173:
```suggestion
If this is your first time using the JIRA CLI, ensure that the config file exists first with `mkdir -p ~/.config/.jira && touch ~/.config/.jira/config.json`. You'll also need to mount the JIRA config file as writeable by setting the `jira_dir_rw: true` configuration (or `export OCMC_JIRA_DIR_RW: true`) the first time. Once you've logged in to ocm-container, run `jira init` to do the initial setup.
```

pkg/engine/engine.go
clcollins: Fixed by 75364708cf18098746a0ea7205de0a2e38f6f16f

pkg/engine/engine.go
clcollins: Fixed by 75364708cf18098746a0ea7205de0a2e38f6f16f

clcollins:
@rendhalver Please test with the latest commit (75364708cf18098746a0ea7205de0a2e38f6f16f) and lemme know if you're good with a `/lgtm`!

That commit changes to use a logger, per @iamkirkbater's suggestion.

I've removed the Draft status - this is ready for real review and merge to the main branch when you all are happy with it.

rendhalver:
Thanks @clcollins!
Fabulous work! Let's get his merged!
/lgtm
:shipit:

PR #268, opened 2024-04-17
check CONTAINER_SUBSYS in cluster-console wrapper

T0MASD:
check for podman in cluster-console wrapper


clcollins:
Fixed by #265

PR #279, opened 2024-05-10
Removes publish All by default and locks to localhost for port binding

iamkirkbater:
During some testing we realized that by using the `--publish-all` functionality we inadvertently open up the exposed port to all other hosts on our local networks, posing a potential security risk.

This changes that functionality so that by default it binds the console port to the local interface, so that you can only access the service running in ocm-container from your local machine and not other hosts on the network.

Effectively changes:
```
FROM:
42ba10b210d4  quay.io/app-sre/ocm-container:latest                        22 hours ago   Up 22 hours   0.0.0.0:43407->9999/tcp    dreamy_blackwell

TO:
5660b923b832  quay.io/app-sre/ocm-container:latest                        9 seconds ago  Up 9 seconds  127.0.0.1:35023->9999/tcp  intelligent_hoover
```

I ssh'd into another machine and curl'd my macbook and it served the contents running on port 43407 to my other machine. After changing the binding to the bottom example, my macbook replied with `curl: (7) Failed to connect to 192.168.1.212 port 35023: Connection refused`

---

As an architecture decision, I figured that it would be better to add a LocalPorts map to the ContainerRef so that we could optionally further expand the availability of ports in the future. For example, in some recent testing it would have been nice to have the ability to have a console running as well as the ability to port forward for PromLens; though that can come in a future PR.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -191,6 +191,10 @@ func (e *Engine) execAndReplace(args ...string) error {
 
 // imageFQDN builds an image format string from container ref values
 func (c ContainerRef) imageFQDN() string {
+	if c.Image.Name == "" || c.Image.Tag == "" {
+		return ""
+	}
+
 	i := fmt.Sprintf("%s:%s", c.Image.Name, c.Image.Tag)
 
 	// The order of the repository and registry addition is important
@@ -264,7 +268,10 @@ func parseRefToArgs(c ContainerRef) ([]string, error) {
 
 	args = append(args, ttyToString(c.Tty, c.Interactive)...)
 
-	args = append(args, c.imageFQDN())
+	imageFQDN := c.imageFQDN()
+	if imageFQDN != "" {
+		args = append(args, imageFQDN)
+	}
 
 	// This needs to come last because command is a positional argument
 	if c.Command != "" {
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -43,6 +43,7 @@ type ContainerRef struct {
 	BestEffortArgs  []string
 	Privileged      bool
 	RemoveAfterExit bool
+	LocalPorts      map[string]int
 }
 
 type VolumeMount struct {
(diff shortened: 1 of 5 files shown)


pkg/engine/engine.go
```diff
@@ -233,8 +234,14 @@ func parseRefToArgs(c ContainerRef) ([]string, error) {
 		args = append(args, "--rm")
 	}
 
+	// If PublishAll is set - publish all of the ports. Otherwise, bind each
+	// one individually to localhost.
 	if c.PublishAll {
```
clcollins, line 239: There are no more references that can reach `PublishAll` anymore, are there?  Should this just be removed?

pkg/engine/engine.go
```diff
@@ -43,6 +43,7 @@ type ContainerRef struct {
 	BestEffortArgs  []string
 	Privileged      bool
 	RemoveAfterExit bool
+	LocalPorts      map[string]int
```
clcollins, line 46: Great idea!  I like the future proofing.

clcollins (changes requested):
This is good, just one suggestion/question.

And if you have time, a test or two would be very welcome. :D

pkg/engine/engine.go
iamkirkbater: I had some time yesterday so I added a hidden flag that emits a warning log if it's used.

pkg/engine/engine.go
```diff
@@ -190,6 +191,10 @@ func (e *Engine) execAndReplace(args ...string) error {
 
 // imageFQDN builds an image format string from container ref values
 func (c ContainerRef) imageFQDN() string {
+	if c.Image.Name == "" || c.Image.Tag == "" {
+		return ""
+	}
```
iamkirkbater, line 196: Had to update the above block because it was _always_ returning `:` if no Image Name or Tag were provided. 

While I understand that's _probably_ okay because they have defaults set during ContainerRef creation as part of the CLI process, this was unexpected from a unit-testing perspective so I just updated this to return an empty string if there's no image references defined.

clcollins:
This looks great, thanks for talking through the changes with me. :shippit:

/lgtm
/approve

PR #282, opened 2024-05-16
Add `--manager`, `--service`, and `--namespace` flags

### LINKED ISSUE #270: Add an option to log directly into a manager or service cluster
As an SRE I'd like to be able to pass a flag such as `--manager` or `--service` to ocm-container to have the logic for logging directly into a service or managment cluster baked into the tooling.

Ideally, if I were to run `ocm-container -C $CLUSTER_ID --manager` it would have the following environment available for me:

* CLUSTER_ID would be set to the CLUSTER ID that was passed
* MC_ID would be set to the Manager Cluster ID
* SC_ID would be set to the Service Cluster ID
* HCP_NS would be set to `ocm-$OCM_ENVIRONMENT-$CLUSTER_ID-$CLUSTER_NAME`
* HC_NS would be set to `ocm-$OCM_ENVIRONMENT-$CLUSTER_ID`
* KUBELET_NS would be set to `kubelet-$CLUSTER_ID`
* oc project would be auto-set to `$HCP_NS`

And if this were a classic ROSA cluster:

* CLUSTER_ID would be set to the CLUSTER_ID that was passed
* HIVE_NAME would be the display name of the hive
* HIVE_NS would be `uhc-$OCM_ENVIRONMENT-$CLUSTER_ID`
* oc project would be auto-set to `$HIVE_NS`

Essentially - the environment variables would be configured in a way similar to if you ran `ocm-container -C $CLUSTER_ID`, loaded into it and then `ocm backplane login $CLUSTER_ID --manager` from within the same container, so all of your env variable context is loaded for $CLUSTER_ID but you are just logged into the Management/Service Cluster.

### LINKED ISSUE #271: Add additional env vars to login
When logging into a cluster I'd like some additional environment variables to be available to me, based on what type of cluster it is, and some for both:

*ALL*:
* OCM_ENV - should be one of `production`, `staging`, `integration`
* OCM_REGION - should default to `global` (this one is for future-proofing, so we can leave this off until we need it)

*Classic*:
* HIVE_NS - should be formatted: `uhc-$OCM_ENV-$CLUSTER_ID`
* HIVE_NAME - should be the user-readable hive name: `hive-stage-01`, etc
* HIVE_ID - should be the Cluster ID of Hive

*Hypershift*:
* MC_ID would be set to the Manager Cluster ID
* SC_ID would be set to the Service Cluster ID
* HCP_NS would be set to ocm-$OCM_ENVIRONMENT-$CLUSTER_ID-$CLUSTER_NAME
* HC_NS would be set to ocm-$OCM_ENVIRONMENT-$CLUSTER_ID
* KUBELET_NS would be set to kubelet-$CLUSTER_ID

clcollins:
Add management/service/hive & namespace login flags

* Adds `--manager` and `--service` flags to allow logging into MC/Hive and SC clusters for a provided `--cluster-id`.
* Adds `--namespace` flag to override the default NS on whatever cluster is being logged into
* Removes login logic from ~/.bashrc.d/* files inside the container, moving the responsibility of login to ocm-container's binary
* Adds related and other helpful ENV vars inside the container

Example ENVs logging into a classic cluster (or it's hive):

```
CLUSTER_NAME=
HIVE_NAME=
HIVE_NS=
CLUSTER_UID=
CLUSTER_ID=
CLUSTER_UUID=
```

And example for HPC:

```
CLUSTER_NAME=
CLUSTER_UID=
SC_ID=
HCP_NS=
CLUSTER_ID=
HC_NS=
MC_ID=
KUBELET_NS=
CLUSTER_UUID=
```

And extra vars that might be helpful, for any cluster:

```
CLUSTER_STS=
CLUSTER_PRIVATE_LINK=
CLUSTER_CCS=
CLUSTER_CLOUD_PROVIDER=
```

Fixes #271
Fixes #270

WIP - TODO:

* Add a log message immediately before spawning login execs so users will have positive feedback on why they're waiting.  This used to be incidental as a result of the different bash scripts, making it feel like login wasn't just hanging, but even though it's the same length if ocm-container executes the login, because there's no feedback it feels like ocm-container has hung.
* Figure out how to perform the login to ocm and backplane in order, but async from the container attachment, to speed up the login.
* Do the plumbing to pick the right cluster if `--manager`, `--service` or `--namespace` are provided.


rendhalver:
I have been thinking about this one.
I think replicating the flags ocm backplane login uses is a good idea so we can just pass them to backplane.
Adding extra flags seems like something we should do in backplane so we aren't doing things differently and we can share idea across the teams.
Other thoughts: does backplane have anything we can use to work out which flags are mutually exclusive so we don't need to keep our own copy of those?

The new --ohss and existing --pd flags would also be good to replicate.
https://source.redhat.com/groups/public/sre/wiki/setup_backplane_cli#isPasted4

Kirk's changes in #295 will also help with the OCM logins.

PR #288, opened 2024-05-30
Replaces UBI with Stream9-minimal image

iamkirkbater:
UBI is not a redistributable image, and should only be used on RHEL machines. Therefore, we should use the CentOS Stream image.

However, the [centos quay repo](https://quay.io/repository/centos/centos?tab=tags) does not tag their images with anything more granular than `stream9-minimal`. This image does look to be updated regularly, though, which may or may not be an issue.

I've built and used this image locally for a few days now without issue on my M1 MBP. 

I've opened this PR for further discussion, and do not have an opinion one way or the other on whether this gets merged or not.

Migrating to stream9-minmal puts us under the "letter of the rule", however since this is an internal tool really used only by RH SREs on CSBs - we are still following the "_spirit of the rule_".

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:9.3-1612
+ARG BASE_IMAGE=quay.io/centos/centos:stream9-minimal
 FROM ${BASE_IMAGE} as tools-base
 ARG OUTPUT_DIR="/opt"
 


iamkirkbater:
:fry-not-sure: - I was told by other engineers on slack that UBI was non-redistributable (and then eventually made this PR), but when I decided to go find a source for that claim somewhere after submitting this PR and googled for it I found the UBI landing page which [explicitly calls out UBI as freely-redistributable](https://catalog.redhat.com/software/base-images#:~:text=The%20Red%20Hat%20Universal%20Base%20Image%20(UBI)%20is%20a%20collection%20of%20Open%20Container%20Initiative%20(OCI)%2Dcompliant%2C%20freely%20redistributable%2C%20container%20base%20operating%20system%20images)...

We can still have the discussion on whether or not to use UBI/stream-9; however I'm perfectly fine with closing this with this new information.... 🤦🏼

iamkirkbater:
More Information: 

From another slack thread - the general consensus is that if the container running UBI as a base is built on a RHEL system it's only allowed to run on that RHEL system. If we built via UBI locally this is fine, as UBI won't have access to any of the RHEL repos, but if being built on a RHEL system it will have the RHEL repos enabled by default, and therefore may not be freely redistributable.

It might just be safer at this point to run Stream9 - though I'm still open to have a discussion.

PR #291, opened 2024-05-31
Install gcloud from backplane-tools

rendhalver:
Install gcloud from backplane-tools


rendhalver:
/hold

Need to make some extra changes.

PR #292, opened 2024-06-06
Update goreleaser/template to v2

clcollins:
Update to use Goreleaser version 2.  I've tested this locally and has been working.  This will require TLs to update their Goreleaser version in use to v2.  Goreleaser will refuse to run v2 templates with v1, so it will be clear to them if they haven't yet updated.


clcollins:
Depends on https://github.com/openshift/backplane-tools releasing v1.1.0++ with the rosa-cli fix.

PR #295, opened 2024-06-29
Allows passing in the ocm config for short-lived ocm token support

iamkirkbater:
Adds initial support for short-lived OCM tokens. You'll need to auth _before_ you get into the container using `ocm login --use-auth-code` and then after you are logged in outside of the container your auth will be passed into the container.

This should remain in draft until the next OCM CLI is released so that we can use the OCM_URL environment variable to override which endpoint to use (int, stage, prod, regional, etc), otherwise this will only work for whatever environment you're logged into OUTSIDE of the conatiner.


rendhalver:
I have tested this with the latest build of our container that has ocm 0.1.74.
Everything worked as expected.

PR #297, opened 2024-07-08
Adds fallback to offline access token login

iamkirkbater:
Adds a fallback for prod access with offline access tokens but adds a warning, and proceeds to use device code login for OCM when in the staging or integration environments.

Per some feedback received in slack this should allow both access mechanisms to give SREs more time to switch without breaking current prod workflows.

### DIFF
--- a/utils/bashrc.d/00-ocm.bashrc
+++ b/utils/bashrc.d/00-ocm.bashrc
@@ -7,7 +7,22 @@ fi
 
 if ! ocm whoami &> /dev/null
 then
-  ocm login --url=$OCM_URL --use-device-code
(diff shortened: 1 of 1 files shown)


iamkirkbater:
I walked through this with an expired OCM token and it worked fine.

But then I logged into OCM outside of the container and attempted to run this again and all logins worked fine.

Then I ran `ocm logout` outside of the container and attempted to run the container again, and now all environments fail to work.

This needs more work and testing before merging.

clcollins:
/hold

Based on @iamkirkbater's last comment

rendhalver:
This is kind of a stopgap feature.
Given that offline tokens will be disabled in Prod soon and this won't be needed after then should we switch focus to getting the new login scheme working nicer?

PR #299, opened 2024-07-08
Adds ability to use GITHUB_TOKEN auth for backplane-tools

clcollins:
Adds the ability to use an exported `GITHUB_TOKEN` env variable for
backplane-tools authenticated installs.


rendhalver:
Holding while we run more builds to test if this fixes some of the rate-limiting.
/hold

PR #300, opened 2024-08-08
Reorder bashrc.d by type; skip automation if launched without bin; fix env and login

clcollins:
This PR reorders and reorganizes the bashrc.d files, placing pure
aliases and functions low (0-9) and labeling them with "libs", exports
and sourcing followign that (10-19) and code execution and automations
after that (20-99). It also removes the shebang (#!) shell declaration
for sourced files, per convention, and sets shellcheck's shell
declaration to allow for continued linting.

This also skips automations (ocm and sre logins) if launching without
the ocm-container binary, to allow the SRE to decide what to do.

Additionally, this fixes sourcing the AWS cli completions, which are
currently broken in this image.

This also updates the way the OCM authentication is handled, moving
authentication to the ocm-container binary. OCM Container will
authenticate with OCM using browser-based auth outside the container if
the user is not logged in, storing the config in ~/.config/ocm/ocm.json
as ocm-cli would.  If the user is already logged in and the tokens are
not expired, OCM Container will use that.  The ocm.json file continues
to be mouted read-only, but the OCM_URL environment variable is set
inside the container to override the OCM URL config setting in the
config file, allowing for OCM Container to be logged into multiple
environments at once, and preventing it from overwriting environments
outside of the container/in other containers. Environment will change
based on the --ocm-url provided (with production being the default).

Finally, this removes to util scripts that have been broken and out of
date since backplane was introduced, and who's functions have largely
been replaced by OCM and Backplane commands, and fixes a single makefile
target that had incorrect syntax.

In practice this should have little impact directly on ocm-contianer
users, other than they will now be able to switch environments just by
setting the config value after login. Login still defaults to
production, and can be overridden by the binary with the OCMC_OCM_URL
env variable that is passed in if the user specifies the `--ocm-url`
flag.

Users using the ocm-contianer binary will stay logged into OCM if their
ocm.json file exists outside the container, so device auth is only
required as their token expires, and would be done manually by the SRE,
if that were to happen.

This is partially in support of
https://issues.redhat.com/browse/OSD-15847, to allow usage of the
ocm-container image elsewhere, and in preparation for changes to the
binary to make device auth login easier.


pkg/engine/engine.go
```diff
@@ -243,7 +243,7 @@ func parseRefToArgs(c ContainerRef) ([]string, error) {
 	if c.PublishAll {
 		args = append(args, "--publish-all")
 	} else if c.LocalPorts != nil {
-		for service, _ := range c.LocalPorts {
+		for service := range c.LocalPorts {
```
clcollins, line 246: This was my a linter change

pkg/ocm/ocm.go
```diff
@@ -18,6 +19,9 @@ const (
 	stagingURL       = "https://api.stage.openshift.com"
 	integrationURL   = "https://api.integration.openshift.com"
 	productionGovURL = "https://api.openshiftusgov.com"
+
+	ocmConfigDest      = "/root/.config/ocm/ocm.json"
+	ocmConfigMountOpts = "rw"
```
clcollins, line 24: RW allows for tokens to be reused until they expire, so you don't have to login every single time

Makefile
```diff
@@ -113,8 +113,7 @@ tag-n-push: registry-login tag push
 
 # Golang-related
 .PHONY: go_build
-go_build:
-	mod fmt lint test build_snapshot
+go_build: mod fmt lint test build_snapshot
```
clcollins, line 116: This was a syntax error, preventing make from parsing this target properly.

pkg/ocmcontainer/ocmcontainer.go
```diff
@@ -125,26 +125,13 @@ func New(cmd *cobra.Command, args []string) (*ocmContainer, error) {
 	maps.Copy(c.Envs, backplaneConfig.Env)
 	c.Volumes = append(c.Volumes, backplaneConfig.Mounts...)
 
-	// Copy the ocm config into the container
```
clcollins, line 128: Moved this into the ocmConfig stuff, to match how the rest of these Envs and Mounts are created, for consistency

utils/bashrc.d/10-completions.bashrc
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+complete -C '/usr/local/aws-cli/aws_completer' aws
```
clcollins, line 3: Fixes the AWS bash completion

utils/bashrc.d/14-kube-ps1.bashrc
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
-source ${HOME}/.bashrc.d/05-kube-ps1.sh
-
-export PS1="[\W {\[\033[1;32m\]${OCMC_OCM_URL}\[\033[0m\]} \$(kube_ps1)]\$ "
+export PS1="[\W {\[\033[1;32m\]\$(ocm config get url)\[\033[0m\]} \$(kube_ps1)]\$ "
```
clcollins, line 2: Ensures the env is properly displayed each time the terminal line updates, so it's always accurate and doesn't require re-sourcing .bashrc

samanthajayasinghe: Can we show short env in PS1`[{prod}]` ? my only concern is that it will show lengthy env like `[{https://api.openshift.com}]`

utils/bashrc.d/20-ocm-login.bashrc
```diff
@@ -0,0 +1,15 @@
...
+  # and allow the user to decide when/if this is necessary
+  return
+fi
+
+if ! ocm whoami &> /dev/null
+then
+  ocm login --url="${OCMC_OCM_URL:-https://api.openshift.com}" --use-device-code
+else
+  ocm config set url "${OCMC_OCM_URL}"
+fi
```
iamkirkbater, line 15: I worry that this will break the ephemeral status of containers. With this change and the read-write nature of the config mount now we're effectively making it so that in each container instance we share state with every other container instance. So if I log into one container in the prod environment and then spin up another in the staging environment this `ocm config set url` command would overwrite the config that's shared between all containers.

What I think might work is the following:

```suggestion
DEFAULT_OCM_URL="https://api.openshift.com"

if ! ocm whoami &> /dev/null
then
  ocm login --url="${OCMC_OCM_URL:-DEFAULT_OCM_URL}" --use-device-code
fi

export OCM_URL="${OCMC_OCM_URL:-DEFAULT_OCM_URL}"
```

Which would prompt the login if the session isn't current and then export the `OCM_URL` environment variable to override all future OCM API calls.

This would still have the side effect of changing the current session outside of the container, though, for SREs who may use OCM outside of the `ocm-container` environment, though I think this is unlikely to happen due to token expiration times and potentially unlikely to cause any major issues since if they were logged out to begin with....

Example scenario:

* SRE is working in the staging env on their workstation, their session expires at some point after the last usage of `ocm`
* SRE opens `ocm-container` in the prod environment to help with a swarm. Because their session is expired it prompts for a login.
* SRE logs in, finishes their task, and exits the container.
* Going back to whatever they were doing, they now though are logged into the `production` OCM environment.

iamkirkbater (commented):
Just some minor thoughts about the drawbacks of making the ocm config readwrite and keeping a global state.

utils/bashrc.d/14-kube-ps1.bashrc
iamkirkbater: We could always overwrite individually in the personalizations mount - IIRC those are automatically sourced at the end of the source change so you could do something like:

```
ocm_short_url() {
  env_url=$(ocm config get url)
  if grep staging <<< $env_url; then
    echo "staging"
    return
  elif grep integration <<< $env_url; then
    echo "integration"
    return
  fi
  echo "production"
 }
 
 export PS1="[\W {\[\033[1;32m\]\$(ocm_short_urll)\[\033[0m\]} \$(kube_ps1)]\$"
 ```
 
 Written entirely from memory but the idea is that would allow you to customize your own PS1 to show whatever you wanted it to.

File: utils/bashrc.d/20-ocm-login.bashrc (hunk 7 above)
Comment (clcollins) [reply utils/bashrc.d/20-ocm-login.bashrc:15]: Ah yeah, I did not consider that case.  It's probably worth just handling the login, then, but that's going to take some more code.

I thought the OCM_URL environment variable was no longer used by OCM, at least the way I was reading the error message that popped up on each login.

Let me slap a hold on this and do some more thinking.

**Comment by clcollins** [comment]
/hold

For some further discussion

File: utils/bashrc.d/20-ocm-login.bashrc (hunk 7 above)
Comment (iamkirkbater) [reply utils/bashrc.d/20-ocm-login.bashrc:15]: IIRC the OCM_URL never was used by OCM before, we only ever used it locally. Now, with the 0.1.74 release of [ocm-cli](https://github.com/openshift-online/ocm-cli/releases/tag/v0.1.74) the OCM_URL environment variable _is_ used by the CLI for all commands except `ocm login`, `ocm config set` and `ocm config get`.

"Deprecating" this from `ocm-container` still makes sense, as we want to move towards prefacing all of `ocm-container`'s environment variables with `OCMC` so I have no issues deprecating the `ocm-container` binary's usage of `OCM_URL`. We would, and should, still use that within the container itself though to control the actions of the `ocm` binary inside the container.

I think some of the other changes in this PR are valuable - what would be the lift to keep all other changes except for the read-write config dir changes and the login flow change? Then we can move this PR forward while we think about the login flow?

File: cmd/configure.go [hunk 8]
```diff
@@ -37,9 +37,9 @@ const (
 const (
 	sshDeprecationMsg                = "SSH multiplexing and Socket mounting is no longer needed or supported. Please remove the 'DISABLE_SSH_MULTIPLEXING' and 'SSH_AUTH_SOCK' fields from your configuration."
 	backplaneConfigDirDeprecationMsg = "The 'BACKPLANE_CONFIG_DIR' field is deprecated and will be removed in a future version. Please remove it from your configuration.  You may specify an alternate backplane config file with 'BACKPLANE_CONFIG'."
-	ocmUrlDeprecationMsg             = "The 'OCM_URL' field is deprecated and will be removed in a future version. Please remove it from your configuration."
+	ocmUrlDeprecationMsg             = "The 'OCM_URL' field is deprecated and no longer used. Please remove it from your configuration."
```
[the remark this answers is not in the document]
clcollins: OCM-Container doesn't currently do anything if you set `OCM_URL=staging ocm-container`.  The environment variable was already deprecated and no longer the way to switch between environments, back when the binary changed to Go.  All the env vars require (and have required) the `OCMC_` prefix.

This change is fully documenting that it's no longer used, and no longer referenced anywhere.

utils/bashrc.d/20-ocm-login.bashrc
clcollins: We can merge the PR with just the change from `rw` to `ro` for the mount options.  Nothing else here will break with that change, except that you'll still be required to auth-in every time you start the container, if you haven't already.  I think that functionality will only impact those of us not using OCM outside the container, though.

cmd/configure.go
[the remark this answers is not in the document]
rendhalver: I have mine set like this:
I think that would still work after this change, I will check when I get a chance.
```alias ocm-container-int='ocm-container --ocm-url=integration'
alias ocm-container-stg='ocm-container --ocm-url=staging'```

File: utils/bashrc.d/14-kube-ps1.bashrc (hunk 6 above)
Comment (rendhalver) [reply utils/bashrc.d/14-kube-ps1.bashrc:2]: Can we get the short url from ocm somehow so we don't need ocm-container specific code to do it?
Or push our changes upstream?

Comment (clcollins) [reply utils/bashrc.d/14-kube-ps1.bashrc]: FWIW, this had to be moved back to ${OCM_URL}, as the `ocm config get url` no longer necessarily gets the correct URL, as we're overriding the ocm config URL with the env var.

This will still be a long-ish URL though, but also no different from what is currently displayed in OCM Container, I believe.

I'm happy for any other commits to make this cleaner in the future though.

File: cmd/configure.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (clcollins) [reply cmd/configure.go:40]: ^ Yeah, that is the current way to do it, since the ocm-container binary rewrite, and should continue to work with this PR.

File: utils/bashrc.d/20-ocm-login.bashrc (hunk 7 above)
Comment (clcollins) [reply utils/bashrc.d/20-ocm-login.bashrc]: I think this is solved.  OCM login occurs outside of the container, the config is updated with the latest tokens, but the URL is not rewritten in the config (on purpose).  This will prevent non-ocm-container OCM results from changing unexpectedly.

Within ocm-container, the results will also not change, but the OCM_URL env is set per container, so multiple environments can be logged in, ephemerally.  

Changing the env inside the container also works, and is not populated back to other containers or the host OCM.

The ocm.json is mounted RO.

**Comment by clcollins** [comment]
/unhold

@iamkirkbater @rendhalver @samanthajayasinghe - I think I've fixed, or explained, all the issues here.  This is ready for a re-review, if you have some time.

I've updated the PR description with how things exist in the latest commit, not the original commit.  Let me know if you have any concerns.

**Comment by rendhalver** [comment]
> @iamkirkbater @rendhalver @samanthajayasinghe - I think I've fixed, or explained, all the issues here. This is ready for a re-review, if you have some time.

Having a look now. )

> It's worth noting that we should cut a new release after this merges, I think, and encourage everyone to use the latest build image, as the internals of the image changed as well as the external binary.

Yep I totally agree with cutting a new release after we merge.

**Comment by rendhalver** [comment]
Testing notes::
I built the image and binary locally and it looks to be all working if I pass in a cluster-id.

```
$ ./build/ocm-container -t latest --registry "" --repository localhost --cluster-id c03103eb-1571-498d-b1fd-70587b445faa
Using config file: /home/pebrown/.config/ocm-container/ocm-container.yaml
[snip]
Checking the context on 15uq6splkva07rae2hgih6eph4vs3p8m
Encountered Errors during data collection. Displayed data may be incomplete: 
	cluster is not a HCP/Management Cluster
=================================================
osd-v4stg-aws -- 15uq6splkva07rae2hgih6eph4vs3p8m
=================================================
[snip]
[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_ID
15uq6splkva07rae2hgih6eph4vs3p8m
[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_UUID
c03103eb-1571-498d-b1fd-70587b445faa
[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_NAME
osd-v4stg-aws
```

I did notice that if we use the cluster name for osd-v4stg-aws it doesn't do all of that but that's because we are searching on the display name in `utils/bashrc.d/06-sre-login-libs.bashrc` and osd-v4stg-aws has it set to this:
`Display Name:		SRE long lived cluster in production: osd-v4stg-aws`
So it's a bit of a unique case.

```
$ ./build/ocm-container -t latest --registry "" --repository localhost --cluster-id osd-v4stg-aws
Using config file: /home/pebrown/.config/ocm-container/ocm-container.yaml
Error: Can't save config file: can't write file '/root/.config/ocm/ocm.json': open /root/.config/ocm/ocm.json: read-only file system
[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_ID

[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_UUID

[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ echo $CLUSTER_NAME


[~ {https://api.openshift.com} (osd-v4stg-aws.z9a9.p1:default)]$ 
```

**Comment by rendhalver** [comment]
I am going to put a hold on this and lgtm it.
Feel free to unhold other people are happy with it.

/hold
/lgtm

**Comment by clcollins** [comment]
> I did notice that if we use the cluster name for osd-v4stg-aws it doesn't do all of that but that's because we are searching on the display name in `utils/bashrc.d/06-sre-login-libs.bashrc` and osd-v4stg-aws has it set to this:
> `Display Name: SRE long lived cluster in production: osd-v4stg-aws`

Ah, good catch @rendhalver !  We're using `display_name` in that query, but most clusters have a longer description in that field.  It looks like ocm-cli actually uses `name` when it does the same lookup:

https://github.com/openshift-online/ocm-cli/blob/main/pkg/cluster/cluster.go#L247

Using "name" instead returns the info for osd-v4stg-aws as expected:  

```
ocm list clusters '--parameter=search=((id like '\''osd-v4stg-aws'\'' or external_id like '\''osd-v4stg-aws'\'' or name like '\''%osd-v4stg-aws%'\''))' --columns 'id, external_id, name' --no-headers
REDACTED  REDACTED  osd-v4stg-aws 
```

That has probably always been broken.  I'm happy to put in a bug fix in a follow up PR, if that's cool with you.

clcollins:
Actually, @rendhalver, I just went ahead and fixed that: https://github.com/openshift/ocm-container/pull/301

rendhalver:
> That has probably always been broken. I'm happy to put in a bug fix in a follow up PR, if that's cool with you.

Oh I am totally happy getting that sorted after this one is merged. :)

PR #303, opened 2024-08-27
 Suppress unnecessary output at normal log levels

clcollins:
Suppress unnecessary output at normal log levels

As an SRE I would like non-debug output suppressed when running
OCM Contianer.

Currently, it prints the container engine image pull
output to the screen. This is messy and not compliant with the "no news
is good news" Unix/Linux terminal/cli philosophy.

This PR has a case switch for log levels > debug, appending
the `--quiet` flag to the Podman/Docker create command, which will
suppress the output for any higher log levels.

OCM Container also prints the config file used for Viper argument
parsing.

This PR also removes that config file parsing ouput, unless there is an
error, since location is standardized and documented.

Secondary commit unrelated to this main PR includes a minor spelling fix to the CLI flags help message.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -183,7 +183,7 @@ var standardFlags = []cliFlag{
 		name:     "publish-all-ports",
 		flagType: "bool",
 		value:    "false",
-		helpMsg:  "Publishes all defined ports to all interfaces. Equivelant of `--publish-all`",
+		helpMsg:  "Publishes all defined ports to all interfaces. Equivalent of `--publish-all`",
 		hidden:   true,
 	},
 }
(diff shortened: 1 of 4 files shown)


cmd/flags.go
```diff
@@ -183,7 +183,7 @@ var standardFlags = []cliFlag{
 		name:     "publish-all-ports",
 		flagType: "bool",
 		value:    "false",
-		helpMsg:  "Publishes all defined ports to all interfaces. Equivelant of `--publish-all`",
+		helpMsg:  "Publishes all defined ports to all interfaces. Equivalent of `--publish-all`",
```
clcollins, line 186: Minor spelling fix.

PR #324, opened 2025-06-16
Various updates for patches/go version

clcollins:
Updates packages and go version for [SREP-867](https://issues.redhat.com//browse/SREP-867)
Reworks OCM API calls to use ocm-common, as part of the above updates
Updates to related to linter fixes
Updates golangci-lint and goreleaser versions
Minor Makefile changes

---

- **Update dependency packages**
- **Update engine tests to sort strings for testing**
- **Fix fmt make target**
- **Update golangci-lint and goreleaser versions**
- **Add fixes for golangci-lint**
- **Fix build-snapshot make target name**
- **Change name of the go-build target to be consistent**
- **Renaming build_binary make target for consistency**


.golangci.toml
```diff
@@ -0,0 +1,8 @@
+# .golangci.toml
```
joshbranham, line 1: nit: Any reason you chose the `toml` format for the config, or can we use the `.yaml` format of the config?

clcollins: No reason.  Fixed.

Also found there's a standard preset for this exclusion, so I switched to using that.

cmd/configure.go
```diff
@@ -299,10 +294,10 @@ func readLegacyConfig() (map[string]any, error) {
 	f, err := os.Open(legacyCfgFile)
 	//lint:ignore SA5001 setting defer before the error check,
 	// because the function will return immediately with the error
```
iamkirkbater, line 296: Do we need to remove these comments with moving the defer below the error check?

cmd/flags.go
```diff
@@ -263,7 +251,11 @@ var disableFeatureFlags = []cliFlag{
 // checks if they are set in viper, and returns an error if they are not.
 func checkFlags(cmd *cobra.Command) error {
 	cmd.Flags().VisitAll(func(f *pflag.Flag) {
-		viper.BindPFlag(f.Name, f)
+		err := viper.BindPFlag(f.Name, f)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error binding flag %s: %v\n", f.Name, err)
```
iamkirkbater, line 256: Is there something the end user should do here? Is this a configuration issue on the user's end or would this indicate a bug in the code? Maybe a better way to ask is - what should an end user _do_ with the information from this error?

cmd/root.go
```diff
@@ -100,7 +100,10 @@ and other Red Hat SRE tools`,
 		o.ExecPostRunNonBlockingCmds()
 
 		if !viper.GetBool("--headless") {
-			o.Attach()
+			err = o.Attach()
+		}
+		if err != nil {
+			return err
```
iamkirkbater, line 106: could we just `return o.Attach()` here? I assume the "attachment" is the last thing we want to do in a command?

pkg/ocm/ocm.go
```diff
@@ -188,8 +193,12 @@ func New(ocmcOcmUrl string) (*Config, error) {
 
 // url takes a string in the form of urlAliases, and returns
 // the actual OCM URL
-func url(s string) string {
-	return urlAliases[s]
+func url(s string) (string, error) {
+	u, ok := urlAliases[s]
+	if !ok {
+		return "", errInvalidOcmUrl
```
iamkirkbater, line 199: While this returns urlAliases, should we also exit at the beginning of this function if the url starts with `http` to allow users to point to a non-aliased OCM URL (or something running locally, etc)? Just trying to think of additional use-cases outside of our specific one that we can "future-proof" for little up-front work :D

cmd/root.go
clcollins: This is specifically a linter fix, but otherwise just as you say.  It's not really useful IRL.

PR #326, opened 2025-06-20
[propose] osdctl needs vault token for accessing dt logs

T0MASD:
osdctl needs vault token for accessing dt logs, those who enable osdctl now also will mount ~/.vault-token when missing an error is thrown:
```
Error: error: problem reading source volume: /home/tomas/.vault-token: stat /home/tomas/.vault-token: no such file or directory        
```
which is rectified by `touch` vault auth

### DIFF
--- a/pkg/featureSet/osdctl/osdctl.go
+++ b/pkg/featureSet/osdctl/osdctl.go
@@ -4,6 +4,7 @@ import "github.com/openshift/ocm-container/pkg/engine"
 
 const (
 	osdctlConfigDir = ".config/osdctl"
+	vaultTokenFile  = ".vault-token"
 )
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


pkg/featureSet/osdctl/osdctl.go
```diff
@@ -14,11 +15,18 @@ func New(home string) (*Config, error) {
...
-	})
-
+	config.Mounts = append(
+		config.Mounts,
+		engine.VolumeMount{
+			Source:       home + "/" + osdctlConfigDir,
+			Destination:  "/root/" + osdctlConfigDir,
+			MountOptions: "ro",
+		},
+		engine.VolumeMount{
```
iamkirkbater, line 25: This will need to be hidden behind a conditional in case the file does not exist on the host system.

I suggest something like the following pseudocode:

```
if os.Stat(home + "/" + vaultTokenFile) {
  config.Mounds = append(
    config.Mounts,
    engine.VolumeMount{
      Source:       home + "/" + vaultTokenFile,
	  Destination:  "/root/" + vaultTokenFile,
	  MountOptions: "rw",
    })
}
```

iamkirkbater (changes requested):

pkg/featureSet/osdctl/osdctl.go
T0MASD: fixed up in [34ed777](https://github.com/openshift/ocm-container/commit/34ed777d25db8328c405eaf39cf031897581a61c)

PR #343, opened 2025-06-30
micro

clcollins:
- **Update docs for running built images**
- **Update base image, omc, jira versions**
- **osdctl needs vault token for accessing dt logs, those who enable osdctl now need to have the token**
- **updates deprecated goreleaser value**
- **fixes build error**
- **Add micro build target**


Makefile
```diff
@@ -54,21 +52,42 @@ checkEnv:
...
 	@${CONTAINER_ENGINE} build $(BUILD_ARGS) -f Containerfile -t $(IMAGE_NAME):$(TAG) --target=ocm-container
+else
+	@${CONTAINER_ENGINE} build $(BUILD_ARGS) --platform=linux/$(ARCHITECTURE) -f Containerfile -t $(IMAGE_NAME):$(TAG)-$(ARCHITECTURE) --target=ocm-container
+endif
 
 .PHONY: build-image-amd64
-build-image-amd64:
-	@${CONTAINER_ENGINE} build $(BUILD_ARGS) -f Containerfile --platform=linux/amd64 -t $(IMAGE_NAME):$(TAG)-amd64 --target=ocm-container
+build-image-amd64: ARCHITECTURE=amd64
+build-image-amd64: build
```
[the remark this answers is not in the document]
clcollins: You're right, I forgot to remove the line below the `build-image-X: build` lines - its not necessary after setting the variable and calling build (which was my attempt to make it more DRY).  Thanks for catching that!

iamkirkbater:
Weird - so it looks like when I ran `.ci/pull-request-check.sh` it failed with the same error locally. I removed the `@` symbols on the build targets to debug, and was greeted with:

```
➜ .ci/pull-request-check.sh
/opt/homebrew/bin/podman build "--no-cache"  "--build-arg GITHUB_TOKEN=[REDACTED]" -f Containerfile --platform=linux/amd64 --target=ocm-container-micro -t ocm-container-micro:amd64 .
Error: unknown flag: --build-arg GITHUB_TOKEN
See 'podman build --help'
make: *** [build-micro] Error 125
```

Thought it was weird that some of these had quotes but others didn't, and when I looked at the Makefile I saw that those two arguments were the only ones that were quoted, so Make was just passing the quotes as literals which was then breaking the script.

I pushed the changes - expecting them to land in my own fork and then I'd make a PR to your branch, but it just pushed right back to your fork I guess haha.

The only other comment I have is can you look at the `.ci/pull-request-check.sh` script and give it a quick once-over to see if there's anything we _want_ to remove there? Probably some duplication now that we're handling this in the Makefile itself.

Overall though, I love this. I'm going to add a hold _and_ an LGTM here. If you want to clean the pull-request-check.sh file up then feel free to do that and I'll give it an LGTM again. Otherwise, just remove the hold and allow it to merge.

/hold
/lgtm

utils/dockerfile_assets/github_dl.py
```diff
@@ -0,0 +1,159 @@
...
+    parser.add_argument("--checksum_file", required=True, help="Name of the checksum file to download (e.g., 'checksums.txt')")
+    parser.add_argument("--checksum_algorithm", default="sha256", help="Checksum algorithm to use (default: 'sha256') ")
+    parser.add_argument("--platform", required=True, help="Platform to download assets for (e.g., 'amd64')")
+    args = parser.parse_args()
+
+    if os.environ.get("GITHUB_TOKEN"):
+        args.token = os.environ["GITHUB_TOKEN"]
+    else:
+        args.token = None
+        print("No GITHUB_TOKEN found in environment variables. Proceeding without authentication.")
```
Mhodesty, line 135: Maybe add a CLI argument that would allow someone to set a specific token file as an argument?

clcollins: I initially had that, but removed it to keep from leaking secrets in, eg, top and ps.  

It _probably_ wouldn't matter.  Almost every case, this script will never be used directly by a person, but I didn't want to leave it in.

PR #351, opened 2025-08-08
update packages and base image

clcollins:
Update multiple go packages.
Update base image to ubi-minimal:10
Replace epel 9 with epel 10 repos
Remove unused rhash package
Add build metadata labels to image builds in Makefile

Both 'go build' and 'make test' pass successfully.


clcollins:
Addressed by #351

/close

PR #369, opened 2025-10-06
initial refactor of features

iamkirkbater:


pkg/features/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,97 @@
...
+		"ro",
+		"rw",
+	}
+	if !slices.Contains(validMountOptions, cfg.Mount) {
+		return fmt.Errorf("invalid mount option. Valid options are %s", validMountOptions)
+	}
+	return nil
+}
+
+func New() (features.OptionSet, error) {
```
iamkirkbater, line 44: Idea here is that each feature independently gets its own config (not passed from the main func) and returns a well defined OptionSet.

Next steps - each feature handles its own "am I enabled" check and its own registration with the features controller.

PR #422, opened 2025-11-07
fix: convert GITHUB_TOKEN ARG to ENV and remove duplicate GH_TOKEN

T0MASD:
## Problem

The `GITHUB_TOKEN` ARG in the Containerfile was never converted to an ENV variable, making it inaccessible to `github_dl.py` and causing all local builds to run unauthenticated and hit GitHub API rate limits (60 req/hour).

Additionally, a duplicate `ARG GH_TOKEN=${GITHUB_TOKEN:-}` existed as leftover code from refactoring that served no purpose.

## Solution

- Remove duplicate `ARG GH_TOKEN=${GITHUB_TOKEN:-}` 
- Convert `ARG GITHUB_TOKEN` to `ENV GITHUB_TOKEN=${GITHUB_TOKEN}` to make it accessible to RUN commands
- Standardize on single `GITHUB_TOKEN` variable

## Verification

Built locally with authentication:

```bash
gh auth login  # Authenticated with classic PAT
GITHUB_TOKEN=$(gh auth token) make BUILD_ARGS="--build-arg BASE_IMAGE=registry.access.redhat.com/ubi9/ubi:9.5" build-full
```

**Results:**

✅ **GitHub API quota check confirmed authentication:**

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -14,9 +14,8 @@ COPY utils/dockerfile_assets/github_dl.py /usr/local/bin/github_dl
 FROM tools-base as backplane-tools
 ARG OUTPUT_DIR="/opt"
 
-# Set GH_TOKEN to use authenticated GH requests
+# Set GITHUB_TOKEN for authenticated GitHub API requests
 ARG GITHUB_TOKEN
-ARG GH_TOKEN=${GITHUB_TOKEN:-}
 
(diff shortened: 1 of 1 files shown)


iamkirkbater:
I've migrated this to utilize podman secrets in #441 so I will be closing this.

PR #427, opened 2025-11-18
Remove pd-cli

clcollins:
Removes PagerDuty CLI as it is no longer maintained and new tooling offers similar features.

### Remove pd-cli tool
- Removed pd-cli binary from container image
- pd-cli is no longer actively maintained
- Users should migrate to alternative PagerDuty tooling

### Update PagerDuty token path
- Changed default token path from `~/.config/pagerduty-cli/config.json` to `~/.config/pagerduty/token.json`
- Updated all code, tests, and documentation to reflect new path
- Simplified directory structure (removed `-cli` suffix)

### Improve security defaults
- Changed default mount option from `rw` to `ro` for better security
- Token file is now mounted read-only by default- Users can still override to `rw` if needed via configuration

### Documentation updates
- Removed "Initial Setup" section from pagerduty.md (no longer needed without pd-cli)
- Updated example_config.yaml with new defaults
- Added migration notes in config_migration.md explaining:
  - New token path
  - Mount option security change
  - pd-cli removal and rationale
  - Feature retention for other PagerDuty tools

## Migration Guide

Users with existing PagerDuty configurations should:

1. Move token file from `~/.config/pagerduty-cli/config.json` to `~/.config/pagerduty/token.json`
2. Update any custom configuration to use new path
3. Note that token is now mounted read-only by default (can override with `config_mount: rw` if needed)

The PagerDuty feature remains available for mounting tokens for use with other PagerDuty tools and integrations.

### DIFF
--- a/docs/config_migration.md
+++ b/docs/config_migration.md
@@ -166,18 +166,70 @@ New configuration options:
 Note: The `config_file` must exist for the feature to work. The `token_file` is optional.
 
 #### PagerDuty
-By default, Pagerduty now looks for the PD token file at `~/.config/pagerduty/token.json`. The following configuration yaml changes have been made:
+The `pd-cli` (pagerduty-cli) tool has been removed from the container as it is no longer maintained. The PagerDuty feature has been updated to allow token mounting for use with other PagerDuty tools.
+
+**Configuration Changes:**
 
 ```yaml
+# Flag changes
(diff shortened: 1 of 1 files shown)


iamkirkbater:
Can we remove the `pkg/features` files for this as well?

iamkirkbater:
Moved to #433

PR #499, opened 2026-04-04
fix: prevent repeated cluster login on unreachable clusters

clcollins:
## Summary
- When a cluster is unreachable, `sre-login` fails but every new bash shell (tmux panes, subshells) retries login automatically, causing the "Logging into cluster..." delay each time
- Uses a `~/.session/sre-login-attempted` marker file so automatic login is only attempted once per container session
- Applies the same once-only treatment to `osdctl cluster context` via `~/.session/osdctl-context-attempted`
- Users can still retry manually with `sre-login $CLUSTER_ID` at any time; successful manual login clears the marker
- Marker files stored in `~/.session/` rather than `/tmp/` for better container engine compatibility

## Test plan
- [ ] Launch ocm-container with a valid, reachable cluster — confirm login works normally and subsequent shells skip via kubeconfig
- [ ] Launch ocm-container with an unreachable cluster — confirm login is attempted once, subsequent shells print skip message instead of retrying
- [ ] After a failed auto-login, run `sre-login $CLUSTER_ID` manually — confirm manual retry works and marker is cleared on success
- [ ] Verify `osdctl cluster context` only runs once per container session
- [ ] Verify `~/.session/` directory is created automatically

### DIFF
--- a/utils/bashrc.d/26-sre-login.bashrc
+++ b/utils/bashrc.d/26-sre-login.bashrc
@@ -4,5 +4,11 @@
 # This prevents another attempt at login if using a terminal multiplexer
 if ! oc config current-context &>/dev/null && [ -z "$SKIP_CLUSTER_LOGIN" ] && [ -n "$CLUSTER_ID" ]
 then
-  sre-login "$CLUSTER_ID"
+  if [ -f "${HOME}/.session/sre-login-attempted" ]; then
+    echo "Skipping automatic cluster login (previous attempt detected). Run 'sre-login $CLUSTER_ID' to retry manually."
+  else
+    mkdir -p "${HOME}/.session"
+    touch "${HOME}/.session/sre-login-attempted"
+    sre-login "$CLUSTER_ID"
+  fi
 fi
--- a/utils/bashrc.d/99-osdctl-cluster-context.bashrc
+++ b/utils/bashrc.d/99-osdctl-cluster-context.bashrc
@@ -1,7 +1,9 @@
 # shellcheck shell=bash
 # show cluster context at login
 
-if [ -n "$CLUSTER_ID" ] && [ -z "$SKIP_CLUSTER_CONTEXT" ]; then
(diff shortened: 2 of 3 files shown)


clcollins:
@clcollins PR #499 has been fully qualified.

**Qualification Summary:**
- **Loops:** 1 iteration of Phases 1–4
- **New commits:** 0 commits added during qualification
- **Merge:** No conflicts (MERGEABLE)
- **CI:** All 7 Konflux checks passing (micro, minimal, full builds + enterprise contract)
- **Copilot:** ⏭️ N/A (not enabled)
- **CodeRabbit:** ⏭️ N/A (not enabled)
- **Codecov:** ⏭️ N/A (not configured)

**Intent validation:** The PR still fulfills its original intent — preventing repeated `sre-login` and `osdctl cluster context` retries on every bash shell when a cluster is unreachable. No changes were needed during qualification.

This PR is ready to merge.

---
*Qualified by Claude on behalf of @clcollins*

utils/bashrc.d/26-sre-login.bashrc
```diff
@@ -4,5 +4,10 @@
 # This prevents another attempt at login if using a terminal multiplexer
 if ! oc config current-context &>/dev/null && [ -z "$SKIP_CLUSTER_LOGIN" ] && [ -n "$CLUSTER_ID" ]
 then
-  sre-login "$CLUSTER_ID"
+  if [ -f /tmp/.sre-login-attempted ]; then
```
iamkirkbater: I've had awful luck in the past with using the tmp filesystem in containers. I've seen both docker on MacOS and podman both struggle here. Not saying this will be a problem right now, but would you be opposed to just putting this in the home directory?

We could even put it in something like `${HOME}/.session/.sre-login-attempted` so that all of these tmp helper files are in a single hidden folder.

iamkirkbater (commented):
Overall love the idea, but would like to discuss the potential of using a `.session` folder in the home directory rather than the /tmp filesystem.

utils/bashrc.d/26-sre-login.bashrc
clcollins: Works for me!  Changed!

clcollins:
@iamkirkbater Mind taking a second look at the new .session stuff?

PR #533, opened 2026-06-11
Re-enable yq; bump backplane-tools to v1.4.0; remove servicelogger

clcollins:
## Summary

- Re-enables `yq` in the minimal container image now that [SREP-1737](https://redhat.atlassian.net/browse/SREP-1737) is complete
- Bumps `backplane-tools` from `v1.2.0` to `v1.4.0`, which includes the arm64 yq fix and is required for yq to be present in the build
- Removes `servicelogger` COPY/RUN lines, as it was dropped from backplane-tools in v1.4.0

## Test plan

- [x] `make test` — all Go unit tests pass
- [x] `make build-image-arm64` — all three image variants (micro, minimal, full) build successfully for arm64, including the yq COPY and `yq --version` smoke test

## Summary by CodeRabbit

* **Chores**
  * Updated backplane-tools to v1.4.0.
  * Modified container image to include and enable the yq tool, removing the previous servicelogger binary.

* **Documentation**
  * Clarified documentation to note the minimal container variant now includes yq instead of the servicelogger.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -14,7 +14,7 @@ COPY utils/dockerfile_assets/github_dl.py /usr/local/bin/github_dl
 FROM tools-base as backplane-tools
 ARG OUTPUT_DIR="/opt"
 
-ARG BACKPLANE_TOOLS_VERSION="tags/v1.2.0"
+ARG BACKPLANE_TOOLS_VERSION="tags/v1.4.0"
 ENV BACKPLANE_TOOLS_URL_SLUG="openshift/backplane-tools"
 ENV BACKPLANE_TOOLS_URL="https://api.github.com/repos/${BACKPLANE_TOOLS_URL_SLUG}/releases/${BACKPLANE_TOOLS_VERSION}"
 ENV BACKPLANE_TOOLS_CHECKSUM_FILE="checksums.txt"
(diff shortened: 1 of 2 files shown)


clcollins:
Addressing the three pre-merge check failures from CodeRabbit — we will not be fixing any of these:

**No-Weak-Crypto**: MD5 is not weak in this context. It is being used solely to validate download integrity against a checksum file provided by the upstream release (o-must-gather). We can only use the algorithm the upstream project actually publishes — we have no control over that. This is not a cryptographic security use case; it is a download integrity check.

**Container-Privileges**: This container is intentionally designed to run as root. It is an interactive SRE toolbox that runs on a user's laptop, not a workload deployed to a cluster. Root access inside the container is required for the tooling to function correctly (cluster credential management, oc, backplane, etc.). This is by design and well understood by the users of this project.

**No-Sensitive-Data-In-Logs**: Printing the first 4 bytes of a token in a local build log is not a meaningful security concern for this project. Builds are run locally by individual SREs, not in shared CI infrastructure where logs would be exposed to untrusted parties. We are not going to fix this.

🤖 Claude <claude@anthropic.com> commenting on behalf of @clcollins

PR #580, opened 2026-07-27
Pin epel-release to Koji build for multi-arch consistency

clcollins:
## Summary
- Pin `epel-release` RPM to a specific Koji build artifact instead of using the floating `epel-release-latest-10` symlink from `dl.fedoraproject.org`
- Prevents Enterprise Contract `rpm_packages.unique_version` failures caused by CDN mirror skew during concurrent multi-arch builds
- Uses a build `ARG` so future version bumps are a one-line change

## Background
The `epel-release-latest-10.noarch.rpm` symlink on `dl.fedoraproject.org` can resolve to different versions depending on CDN mirror state (`10-8.el10_2` vs `10-8.el10_3`). When x86_64 and arm64 builds run concurrently, they can install different versions of the same noarch RPM, causing the EC `rpm_packages.unique_version` rule to fail.

Koji (`kojipkgs.fedoraproject.org`) serves individual build artifacts directly and is not subject to mirror propagation delays. The RPM is GPG-signed with the EPEL key regardless of download source.

## Test plan
- [ ] Multi-arch container build succeeds for all three variants (micro, minimal, full)
- [ ] Enterprise Contract `rpm_packages.unique_version` check passes
- [ ] EPEL packages (e.g., `fzf`) still install correctly in the full image

## Summary by CodeRabbit

* **Chores**
  * Updated the container build to install a pinned EPEL release package instead of a dynamically selected “latest” release.
  * Improved build consistency and predictability by avoiding runtime detection of the platform version.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -135,10 +135,6 @@ ENV IO_OPENSHIFT_MANAGED_COMPONENT="minimal"
 ### DNF Install other tools on top of Minimal
 FROM ocm-container-minimal as dnf-install
 
-# Add epel repos - dynamically detect RHEL version
-RUN RHEL_VERSION=$(rpm -E %{rhel}) && \
-    rpm --import https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-${RHEL_VERSION} && \
-    rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-${RHEL_VERSION}.noarch.rpm
 
 # Install packages
 # These packages will end up in the final image


clcollins:
/override "Red Hat Konflux / ocm-container-main-enterprise-contract / ocm-container-micro" "Red Hat Konflux / ocm-container-main-enterprise-contract / ocm-container-minimal"

The micro/minimal EC checks evaluate a snapshot that includes the full ocm-container image from the previous master build, which still has EPEL installed (causing rpm_packages.unique_version skew between arches). The full image EC check and the pr group EC check both passed with 0 failures on this PR, confirming that the EPEL removal in the newly-built full image resolves the issue. Once merged, subsequent master builds will produce clean snapshots for all components.

🤖 Claude <claude@anthropic.com> commenting on behalf of @clcollins

